PR #5, opened 2018-01-07
Code cleanups

Broxzier:
This PR fixes some warnings, adds comments at the end of namespace braces, makes all source files have a EOL at the end of the file (strictly a standard, but compilers don't really care), and removed unnecessary checks for `nullptr` before `delete`ing.

You'll notice that the diff for the files `LinuxDolphinProcess.h`, `LinuxDolphinProcess.cpp`, and `MemoryScanner.cpp` is huge, that's because they were commited to the repo with CRLF endings, which I have normalized to LF now.


aldelaro5:
I will have to investigate why the line endings weren't enforced by clang-format because normally, they should be so it's a bit weird.

PR #6, opened 2018-01-09
Visual Studio Project Files

Broxzier:
This PR adds a solution file for Windows contributors. I've only tested it in VS2017 (which already has CMake support), but it should work in older versions with minimal changes.
Even though CMake can generate solution and project files for Visual Studio, personally I prefer a solution file that is not crafted by CMake.

**Pros**
 - Standard VS environment (this also affects plugins, like Visual Assist)
 - Better organized than project files generated through CMake
 - Allow compiling a single file (useful for when refactoring for example)
 - Friendly for new Windows contributors

**Cons**
 - Files generated with Qt's MOC tool have to be present in the project file
 - For other versions of Qt, QTDIR has to be set manually (have to do something similar for the native CMake support as well, so it's not a pro nor a con really).

I'm not entire sure the selected Windows SDK (10.0.16299.0) is installed with VS by default. I know 8.1 isn't. This can be updated by right-clicking the project inside VS, then Retarget Projects.

Something I'd like to improve but would like your input on beforehand is the way Qt's MOC tool is called for the widget files. [It has a lot of duplicated settings now](https://github.com/Broxzier/Dolphin-memory-engine/blob/8865974bd468c1806b2db282503b2f37a61b4735/Source/Dolphin-memory-engine.vcxproj#L137-L184). Would you mind it if I rename all Qt headers to something the MSBuild can recognize easily? An extension like `*.qt.h` or `*.hpp` would work, or maybe `*_qt.h`.

I've tried to make the QTDIR variable editable from within VS inside the project properties page, which was possible with property sheets, however when changed, the property sheet would be updated with the new value. I'd rather see this write the changed values inside the vcxproj.user file. It's how other user settings, like `LocalDebuggerEnvironment` and `ShowAllFiles`, are saved too.

### DIFF
--- a/Source/Dolphin-memory-engine.vcxproj
+++ b/Source/Dolphin-memory-engine.vcxproj
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="Dolphin-memory-engine.default.props" />
   <ItemGroup Label="ProjectConfigurations">
     <ProjectConfiguration Include="Debug|x64">
       <Configuration>Debug</Configuration>
@@ -18,6 +17,7 @@
     <WindowsTargetPlatformVersion>10.0.16299.0</WindowsTargetPlatformVersion>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
+  <Import Project="Dolphin-memory-engine.default.props" />
   <PropertyGroup Condition="'$(Configuration)'=='Debug'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
@@ -36,21 +36,13 @@
   </ImportGroup>
   <ImportGroup Label="Shared">
   </ImportGroup>
-  <ImportGroup Label="PropertySheets" Condition="'$(Configuration)'=='Debug'">
-    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
-  </ImportGroup>
-  <ImportGroup Label="PropertySheets" Condition="'$(Configuration)'=='Release'">
+  <ImportGroup Label="PropertySheets">
     <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
   </ImportGroup>
   <PropertyGroup Label="UserMacros" />
-  <PropertyGroup Condition="'$(Configuration)'=='Debug'">
-    <OutDir>$(SolutionDir)bin\</OutDir>
-    <IntDir>$(SolutionDir)obj\$(Configuration)\</IntDir>
-    <TargetName>$(ProjectName)d</TargetName>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)'=='Release'">
-    <OutDir>$(SolutionDir)bin\</OutDir>
-    <IntDir>$(SolutionDir)obj\$(Configuration)\</IntDir>
+  <PropertyGroup>
+    <OutDir>$(ProjectDir)bin\$(Configuration)\</OutDir>
+    <IntDir>$(ProjectDir)obj\$(Configuration)\</IntDir>
   </PropertyGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)'=='Debug'">
     <ClCompile>
@@ -58,12 +50,12 @@
       <Optimization>Disabled</Optimization>
       <SDLCheck>true</SDLCheck>
       <ConformanceMode>true</ConformanceMode>
-      <AdditionalIncludeDirectories>$(QTDIR)\include;$(QTDIR)\include\QtCore;$(QTDIR)\include\QtGui;$(QTDIR)\include\QtWidgets</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>$(QTDIR)include;$(QTDIR)include\QtCore;$(QTDIR)include\QtGui;$(QTDIR)include\QtWidgets</AdditionalIncludeDirectories>
       <ObjectFileName>$(IntDir)%(RelativeDir)</ObjectFileName>
       <PreprocessorDefinitions>WIN32;WIN64;QT_DLL;QT_CORE_LIB;QT_GUI_LIB;QT_WIDGETS_LIB;%(PreprocessorDefinitions)</PreprocessorDefinitions>
     </ClCompile>
     <Link>
-      <AdditionalLibraryDirectories>$(QTDIR)\lib;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
+      <AdditionalLibraryDirectories>$(QTDIR)lib;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
       <AdditionalDependencies>qtmaind.lib;Qt5Widgetsd.lib;Qt5Cored.lib;Qt5Guid.lib;%(AdditionalDependencies)</AdditionalDependencies>
       <SubSystem>Windows</SubSystem>
     </Link>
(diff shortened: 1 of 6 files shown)


aldelaro5:
I was actually trying to have a solution to work, but after a while, I gave up because it was being really werid with Qt and moc, but since this looks like it solves these problems, I guess I can test it on my vm later,  It wouldn't hurt to add it either because you could still do it the cmake way.

About the headers, do you mean the one inside the source/GUI folder?

As for QTDIR, tbh, the best way would be to have a refference within the git repos to a qt libs repos which means the version of the lib won't have to be installed and you also get to have a minimalistic lib (the project only uses QtCore, QtGUI and QtWidgets).  The problem is I have no idea how to do this and I am really not sure how you are supposed to set that up so I didn't necessarly went for it while there was a solution available.  If you do know how to set this up, please let me know because that would be nice.

Broxzier:
Yes, I mean the ones that MOC has to run over before compiling, namely:
 - `GUI\MainWindow.h`
 - `GUI\MemScanner\MemScanWidget.h`
 - `GUI\MemScanner\ResultsListModel.h`
 - `GUI\MemViewer\MemViewer.h`
 - `GUI\MemViewer\MemViewerWidget.h`
 - `GUI\MemWatcher\DlgChangeType.h`
 - `GUI\MemWatcher\MemWatchModel.h`
 - `GUI\MemWatcher\MemWatchWidget.h`
Actually thinking about it now, they all start with GUI/, which could be used for the selector. Only there is one file that MOC doesn't need to run over: `GUICommon.h`. As long as that's the only one, an exception can be added easily without creating too much of a mess.

Sadly I don't know how to set up git dependencies, but I just found that it can be set from within Visual Studio quite easily with the Qt VS Tools plugin. Once I get something working, I'll add it to the readme file too.

aldelaro5:
yeah I would actually prefer that rather than renaming the headers.

As for  the qt repos, the reason I mention this is because I know Dolphin does that and it has been working quite well because you don't need to do anything unlike this repos where the install is quite a lot tbh.  I think I do have to create a repos on github for this, but I am not sure how the referencing part or the includes would go.

aldelaro5:
update: I searched a bit and the referencing part seems simple enoguh from what I can read here:

https://git-scm.com/book/en/v2/Git-Tools-Submodules

there's the building part, but that I can try to figure out later, especially after this pr because it will have a working sln.

So here's what I'll do: I will have this pr be okay as it is with the QTDIR and test it to make sure everything builds and in case I merge it, I'll try to figure out the qt dependency part, basically, it involves me creating a new repos and then adding a submodule in the main repos so it could be fetched for everyone and then setting vs to include the libs withn the subfolder in the include paths.  Or I could just change the QTDIR from your pr and set it so it works with the fetched repos, whatever, it's details.

Now that I think about it, I will do this before I release 0.4.0 because it will jsut make everything easier from now on, my Qt installs on my vm takes way too mcuh disk space so that would be nice.

Source/Dolphin-memory-engine.vcxproj
```diff
@@ -0,0 +1,189 @@
...
+  </ImportGroup>
+  <ImportGroup Label="PropertySheets" Condition="'$(Configuration)'=='Debug'">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <ImportGroup Label="PropertySheets" Condition="'$(Configuration)'=='Release'">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <PropertyGroup Label="UserMacros" />
+  <PropertyGroup Condition="'$(Configuration)'=='Debug'">
+    <OutDir>$(SolutionDir)bin\</OutDir>
```
aldelaro5: I would prefer here to have the configuration name be a directory too after bin, it does seem to add a d at the end for debug, but I rather have the directory just so it's clear which config the binary was in.

Source/Dolphin-memory-engine.vcxproj
```diff
@@ -0,0 +1,189 @@
...
+    <IntDir>$(SolutionDir)obj\$(Configuration)\</IntDir>
+    <TargetName>$(ProjectName)d</TargetName>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)'=='Release'">
+    <OutDir>$(SolutionDir)bin\</OutDir>
```
aldelaro5: same here

Broxzier:
I'm having issues with the DLL's from the submodule. Compiling works fine, but when I try to run the executable, I get an error. I double checked the DLLs from the submodule with `dumpbin /headers` and even with `file` from cygwin to make sure they are all 64-bit versions. When I then compared my DLLs with the ones in the repo, I got this interesting output:

```
for %F in ("Qt5*.dll") do fc /b %F ..\bin\%F
Comparing files Qt5Core.dll and ..\BIN\QT5CORE.DLL
002C6602: 51 35
002C6603: 74 2E
002C6604: 35 31
002C6605: 2E 30
002C6606: 31 2E
002C6608: 2E 5C
002C6609: 30 6D
002C660A: 5C 73
002C660B: 35 76
002C660C: 2E 63
002C660D: 31 32
002C660F: 2E 31
002C6610: 30 37
002C6611: 5C 5F
002C6612: 6D 36
002C6613: 73 34
002C6614: 76 00
002C6615: 63 00
002C6616: 32 00
002C6617: 30 00
002C6618: 31 00
002C6619: 37 00
002C661A: 5F 00
002C661B: 36 00
002C661C: 34 00
Comparing files Qt5Cored.dll and ..\BIN\QT5CORED.DLL
0071F8E2: 51 35
0071F8E3: 74 2E
0071F8E4: 35 31
0071F8E5: 2E 30
0071F8E6: 31 2E
0071F8E8: 2E 5C
0071F8E9: 30 6D
0071F8EA: 5C 73
0071F8EB: 35 76
0071F8EC: 2E 63
0071F8ED: 31 32
0071F8EF: 2E 31
0071F8F0: 30 37
0071F8F1: 5C 5F
0071F8F2: 6D 36
0071F8F3: 73 34
0071F8F4: 76 00
0071F8F5: 63 00
0071F8F6: 32 00
0071F8F7: 30 00
0071F8F8: 31 00
0071F8F9: 37 00
0071F8FA: 5F 00
0071F8FB: 36 00
0071F8FC: 34 00
Comparing files Qt5Gui.dll and ..\BIN\QT5GUI.DLL
FC: no differences encountered

Comparing files Qt5Guid.dll and ..\BIN\QT5GUID.DLL
FC: no differences encountered

Comparing files Qt5Widgets.dll and ..\BIN\QT5WIDGETS.DLL
FC: no differences encountered

Comparing files Qt5Widgetsd.dll and ..\BIN\QT5WIDGETSD.DLL
FC: no differences encountered
```

So it looks like the `Qt5Core.dll` and `Qt5Cored.dll` seem to differ. When I replace those two DLLs with the ones for my installation, everything works as expected. Any idea how that happened?

aldelaro5:
Checking further, the diff appears to be a string of where the libs would be but the one on the right feels a bit strange (it's missing some characters).

Maybe it got corrupted or something, I will check that on my vm later today, this shouldn't prevent execution I think so I find that strange too.

Btw, just a reminder to address the comments about the output directory of the binaries.

Broxzier:
- Qt5.1.0\5.1.0\msvc2017_64
- 5.10.\msvc217_64

Both seem strange values to me, they're missing path seperators and the versions numbers don't match.

I'll just keep using my DLLs until we know what the issue is, so I can continue working on it. The last commit fixed the output paths. I'm now going to look into how I can reduce the duplicated lines at the end, but that's not really required for this PR.

aldelaro5:
Have you tried to move the dll into the binary folder THEN run it?

You have to do this for the exe to work, cmake had a rule to do this, but not for the sln.

Copy the dll ending by d in debug config or without the d in release config.  You could add the rules in the sln, but you have to do this, I actually do this when distributing releases and it's just me copy pasting.  Copy the one from Externals to test.  If that doesn't work, I'll look into it.

aldelaro5:
I found out that you are supposed to keep the qt_prfpath as it is, but use the deployment tool when doing a release, which is something I haven't done yet.

Basically, the error you got has nothing to do with the path, I even moved my install Qt folder to make sure it would not see them and your sln worked fine with the externals one, you don't need to copy them to make the program run in vs likely because your sln tells to include the dll from the externals.  You would need to do that and other things if you were to run it without vs (in that case, you could use the dployment tool).

I added the deployment tool to the submodule repos as it would help me doing a release, but I won't be changing the dll, I preserved the directory structure from a Qt install which might be why the path was slightly different.

I DID however noticed 2 issues with the sln, first, the path where the sln is placed cannot contain spaces or the mocing fails.  Lastly, don't put the bin folder in the root directory, it's easier if it's in the source folder.

Other than that, you should be able to build AND run within vs,  Try to move entirely your Qt folder just to see if it will just use the Externals one, it should build and run fine.

Once you addressed the bin folder issue (I don't know if you can for the moc one, at least attempt it), I would be okay to merge it,

Broxzier:
That's not entirely true, the path is used when it can't find the DLL it's looking for. The error message I'm getting is: `This application failed to start because it could not find or load the Qt platform plugin "windows" in "".` When I renamed my Qt installation, it suddenly also couldn't find the plugin it was looking for with my version of the Qt5Core.dll. Then I found out that copying the plugins folder to the output directory works too. That means that for some reason setting `PATH` to the plugins folder doesn't work... Copying over the `Submodule/..../bin` and `Submodule/..../plugins` folders to the output direcotry works too. I'm doing that as a build step now, and things work. It's dazes me that setting `PATH` doesn't work.

If you don't dislike the xcopy post-build step, then I will push that. Meanwhile I'll try to figure out what's wrong with how `PATH` is set.

Btw, I think the dependencies repository is missing the styles plugin, because the window looks odd without it.

Without styles:
image
With styles:
image

aldelaro5:
indeed, I got that error too and forgot to add the styles, thanks for reminding me.

As for the qwindows, I retested to be sure and it appears I was wrong, you do have to copy them in.  So yeah, copying the plugin folder and the bin seems like a good idea, I do my releases this way anyhow, I think the only thing you have to make sure is that the debug version are copied for debug config and idem for the release.

I will now add the styles to the qt repos and post a comment when done, you could then test it again.

aldelaro5:
Ok styles are in, there is again a debug and a release version.

Broxzier:
After some more testing, it looks like the plugins folder _needs_ to be where the executable is. Without it, the program just can't find the DLLs, not even when the plugins\\.. folder is in `PATH`, or when the DLLs are in the same directory as the executable. Copying them over looks like the only nice solution here. I'll quickly try getting this done via a target, otherwise just as post-build command line using copy/xcopy.

Broxzier:
The target has been created. I accidentally commited the change made for #6 which I needed here. I'll remove that, and will fix the issue when there are spaces in the paths.

Broxzier:
Okay, now a target has been added that copies over the required dlls. I've fixed the remaining issues you mentioned, and rebased everything on the latest master branch. It's ready to be merged for me.

aldelaro5:
Unfortunately, I can't merge it because after some final tests, it seems to compile fine, but it fails at linking the moc:

https://pastebin.com/NzTgn7JD

It never did this error before so it might be due to change you did today.

Broxzier:
Just a guess here, but it looks like it's linking to multiple versions of the generated cpp files. Do you maybe still have ones that were generated before? If so, try running `git clean -fdx` (add `--dry-run` to see what gets removed first), and then compile again.

aldelaro5:
Yep, that was the problem, huh, it's weird considering I can't really locate the older files, maybe I missed them or something.

PR #7, opened 2018-01-11
Update submodule to include new files

Broxzier:
The submodule was still pointing to an older commit (https://github.com/aldelaro5/win-qt5-dolphin-memory-engine/commit/75f5812eed3b3950d4a5a7e3742c9c8d2c46d42c). This PR updates the submodule the include the new files. You'll have to run `git submodule update` after checking out or merging this branch.

Edit: Just to clarify, because GitHub displays it a bit oddly. This PR doesn't add the new files to this repo, but it updates the submodule which points to the newer commit. You can check what it's currently pointing at by opening the file `.git/modules/Externals/Qt/HEAD`. You'll see it contains the commit hash to an older version. All this commit does is basically setting it to a newer commit hash.


aldelaro5:
oh yeah forgot that you do have to do this.

PR #9, opened 2018-01-16
Improve existing documentation

Ikaheishi:
I took another pass over the readme and made several minor changes all over to further improve what it says. Also fixed a few spelling and grammatical errors.


README.md
```diff
@@ -41,16 +41,16 @@ The compiled binaries should be appear in the directory named `build`.
 ## General usage
 First, open Dolphin and start a game, then run this program. Make sure that it reports that the Wii-only extra memory is present for Wii games and absent for GameCube games.
 
->_The Wii-only extra memory is an extra memory region exclusive to the Wii. Its presence affects what are considered valid watch addresses as well as where the scanner will look._
+>_The Wii-only extra memory mentioned here is an extra memory region exclusive to the Wii known as MEM2. Its presence affects what are considered valid watch addresses as well as where the scanner will look._
```
aldelaro5: I removed all GUI refferences of "MEM2" because it could have been confusing so I don't think it should be refferenced in the documentation, refer to the screenshot.

aldelaro5: The reality is that this is not known much :) you have had to have been working on dolphin or have a relatively good knowledge of the hardware to know the name of the memory region.

I am only ok to keep it in the code because if you work on this, you should be aware of this, but for the average user, most will just not know the name, the only thing NEEDED to know anyway is that this is a region exclusive to the Wii.

Ikaheishi: Alright, so it is not really readme worthy at all then. (Probably only worth an index cross-reference in some big manual thing.)

I've removed the reference and rephrased that paragraph to be cleaner and flow better. _Might_ be a little bit of purple prose, but it sure does look nice and informative. 😃

PR #10, opened 2018-01-22
Simplify user-facing references to the Wii's extended memory segment

Ikaheishi:
> “Regardless, I don't think I did a good job on that paragraph. I already have an idea for a much better rewrite.” — Me, on my previous pull request #9  

So here's part two to my idea for a better rewrite as I mentioned to @aldelaro5 before. “Wii-only extra memory” is a mouthful and kinda awkward, so I sought to come up with a better solution on how to refer to it.

Then there's how it's represented in the GUI. The two states are currently presented with longer strings of text that are only 7.69% different; _way_ too easy to misread in my opinion. And so I also set out to improve that situation too.

My end result is this proposal to start calling it the active “memory type”. This makes things short and sweet, but also differentiates the GUI representation much more. Take a look and compare them for yourself:

**Before:**
```
The extended Wii-only memory is present
The extended Wii-only memory is absent
```

**After:**
```
Memory type: Wii
Memory type: GameCube
```

This way the user's attention is drawn more towards the information we want to convey and they are much more unlikely to misread and confuse which state is active. The change also helps streamline the topic of MEM2's presence (or lack thereof) in discussion and consequently in documentation as well.

Instead of saying “make sure the Wii-only extra memory isn't there when you're doing GameCube stuff” one can just say “make sure the memory type matches what you're working with”, _if it's even necessary to tell someone that._ The change should also make it much more intuitive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,9 +39,9 @@ The compiled binaries should be appear in the directory named `build`.
 
 
 ## General usage
-First, open Dolphin and start a game, then run this program. Make sure that it reports that the Wii-only extra memory is present for Wii games and absent for GameCube games.
+First, open Dolphin and start a game, then run this program. Make sure that it reports that the memory type matches the system being emulated; Wii for Wii games and GameCube for GameCube games.
 
->_Exclusive to the Wii, this is an extra region of memory added as part of the enhancements from the earlier GameCube hardware. Consequently, the presence of this extra memory affects what is considered a valid watch address, as well as what areas of memory the scanner will look though._
+>_Added as part of the enhancements from the earlier GameCube hardware, the Wii has an extra region of memory onboard. Consequently, the presence of this extra memory affects what is a valid watch address and therefore what regions of memory the scanner will look though._
 
 If the hooking process is successful, the UI should be enabled, otherwise, you will need to click the hook button before the program can be of any use.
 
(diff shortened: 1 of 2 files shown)


aldelaro5:
Sorry for the delay on this one, but after the recent release, I needed a little break :)

I thought about it and I don't think it's a good idea.  The reasoning behind is that it hides the fact there's 2 distinct memory regsions.  Now, in the context of the watcher and scanner, this is fine, it's abstracted as one contiguous block, but where it starts to get confusing is the viewer.  The way the viewer works is I have to show both regions separately and you have to browse them separately,  I even named MEM1 as "Common RAM" because it's common to both consoles.  

Another thing to consider is because they are different regions, appart from one another, it could potentially add meaning to the memory.  For example, something being in MEM2 might not be in MEM1 for a reason, maybe the data is particularly big.  This is why I made the viewer this way.

Ultimately, what I want is the user to be aware that MEM2 is present and separated from MEM1.  These changes makes it so that the former is respected, but not the later which might be a problem considering the aforementioned 2 reasons.  I am open to abstract the details that are unecessary to the users (such as endianness, Dolphin's process and their addresses, etc...), but I feel this detail is necessary to present.  This is why btw I renamed the name in the UI, the name MEM2 is confusing and giving a name that tells what it is, a region only present on a WIi, is enough.

aldelaro5:
As discussed over discord, I agreed to rework the paragraph without the notion of the memory type.  This PR will stay open until the paragraph rework is pushed.

aldelaro5:
Can I get a status on this PR?

Ikaheishi:
Sorry… Life happened. Closing this myself until I catch back up on…well…everything. 😔

PR #12, opened 2018-01-27
Remove MFP casts from connect calls

Broxzier:
The static_casts are bloating the code, making it unnecessarily hard to read. The templated connect functions from Qt are smart enough to find the correct MFP.

Only when a signal has multiple overloaded possiblilities, the compiler can't select one, but for this Qt has a helper class `QOverload<>`, in which you only have to provide the argument types to let it pick the correct one. Using this rids the code of the return type and duplicated class name you see in static_cast.

C++14 has an even more elegant solution, but I'm not sure if VS2015 supports this, and it would only remove the `::of` part.

### DIFF
--- a/Source/GUI/MainWindow.cpp
+++ b/Source/GUI/MainWindow.cpp
@@ -50,7 +50,7 @@ void MainWindow::makeMenus()
   connect(m_actImportFromCT, &QAction::triggered, this, &MainWindow::onImportFromCT);
   connect(m_actExportAsCSV, &QAction::triggered, this, &MainWindow::onExportAsCSV);
 
-  connect(m_actViewScanner, static_cast<void (QAction::*)(bool)>(&QAction::toggled), this, [=] {
+  connect(m_actViewScanner, &QAction::toggled, this, [=] {
     if (m_actViewScanner->isChecked())
       m_scanner->show();
     else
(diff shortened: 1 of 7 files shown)


Broxzier:
(not directly related to this PR, but I don't know a better place to ask you this)

Most of the slots have names like `onAddWatchEntry`, and to know what is happening there exactly, you have to look at the code inside the slot. I'd like to improve this by giving slots names with verbs, for this example something like `openWatchDialog`. Besides, I find the "on" prefix confusing, as this is often used in event names (much like signals). 

Would you mind if I change them, or would you like to keep them as is?

aldelaro5:
I would rather keep them simply because they are triggered from events, it's just it's telling what would happen when someone press on a button for example.  As far as I'm aware, most GUI API I used do work like this (Java, winform, WPF, etc...).  Even their defaults are set to on followed by the name of the coresponding widget followed by the event name.

Broxzier:
I understand you want to make it clear that those functions are bound to events, but Qt provides a macro `slots` for this. In the header, you just use `private slots:` for the block with functions that are connected to events. That way you can name the functions after what they do, rather than what they are bound to.

> As far as I'm aware, most GUI API I used do work like this (Java, winform, WPF, etc...).

I don't know about WPF, but the other ones are following the same approach where the events (signals) have the On prefix, and the event handlers (slots) are bound to them. In Java (And JavaScript), you use `Object.AddEventListener(onEventname, myFunction);`, and in WinForm you subscribe to an event by doing `onEventname += myFunction;`

> Even their defaults are set to on followed by the name of the coresponding widget followed by the event name.

Do you mean in Qt? I've not worked much with Qt myself, but looking through their documentation and  header files I mostly see slots having verbs for their names.

aldelaro5:
ok, you can open a PR with that, I'll see which names I am okay with and which one I am not.

PR #13, opened 2018-01-31
Apply Qt coding conventions to includes

Broxzier:
Following https://wiki.qt.io/Qt_Coding_Style and https://wiki.qt.io/Coding_Conventions

One important thing noted in the second page:

> The library prefix is necessary for Mac OS X frameworks and is very convenient for non-qmake projects.

From a quick look it doesn't look like the CMakeList.txt file has to be updated, but I currently don't have a Linux machine or VM available to confirm this.

### DIFF
--- a/Source/CheatEngineParser/CheatEngineParser.h
+++ b/Source/CheatEngineParser/CheatEngineParser.h
@@ -1,8 +1,8 @@
 #pragma once
 
-#include <QFile>
-#include <QString>
-#include <QXmlStreamReader>
+#include <QtCore/QFile>
+#include <QtCore/QString>
+#include <QtCore/QXmlStreamReader>
 
 #include "../MemoryWatch/MemWatchTreeNode.h"
 
--- a/Source/Dolphin-memory-engine.vcxproj
+++ b/Source/Dolphin-memory-engine.vcxproj
@@ -50,7 +50,7 @@
       <Optimization>Disabled</Optimization>
       <SDLCheck>true</SDLCheck>
       <ConformanceMode>true</ConformanceMode>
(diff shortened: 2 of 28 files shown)


aldelaro5:
I do not like the new includes name and I do doubt they are required on mac OS because it seems Dolphin uses the same one I do and I haven't seen build errors on it yet.

The reason I do not like them is it's too explicit, I don't necessarly want to know which libs the file is from where I get the option to just say the name.  For this project, it's fine with only 3 libs from Qt, but even then, I don't like the additional length for readability.

Also, I am primarly developping on an Arch Linux machine using a Windows 10 VM with KVM and both never had problems.  As for mac OS, tbh, I likely won't get a testing setup to add support for it (mainly because of the required price and the fact I can't just have a VM run on my machine), but if someone were to ever add support for it, then I will see if I need to change the includes name, but for now, they are fine as is.  Adding mac OS support is possible, but isn't really a necessity because mac OS isn't a very good platform to run Dolphin in (mainly graphics API issues and other things).

As for the rest, seems fine if it still builds which it probably does.

Broxzier:
Ignoring the potential mac OS port, there's still the advantage that when the includes are being this explicit, includes from the same library are grouped together by clang-format, and the advantage that having the paths to the include directories removed cleans up some of the commands inside the project file.

These are just things I consider cleanups, but if you prefer how it currently is, feel free to close this PR.

aldelaro5:
I wouldn't close it because the changes other than the includes interests me.

Broxzier:
I'm not sure what other changes you mean. The changed settings in the project file are related to the search paths for includes.

aldelaro5:
oh nvm I am dumb, well closing this now.

PR #14, opened 2018-01-31
AppVeyor Build System

Broxzier:
[AppVeyor](https://www.appveyor.com/) is a continuous integration service/build system, which can be used for free for Open Source projects. It only provides VMs for Windows builds, for Linux or other platforms you'll have to use something another CI, like [Travis](https://travis-ci.org/). AppVeyor can also be used to perform unit tests, which might be desirable to add later on (QTest seems interesting).

The file `appveyor.yml` contains the settings for AppVeyor to run. I've set it up to build both the Debug and Release configurations using MSVC 2017, and when successful, it'll create an archieve with the output (executable, DLLs and symbols). Those files should run out-of-the-box.

Before AppVeyor will build this repository, the project has to be added to first. I can only do this for my own fork, so you'll have to do that first. You can log in with your GitHub account, then click Add Project and select this repository. It should only take a minute.

AppVeyor's integration with Github will show a green checkmark for branches and pull requests that build successfully, and a red cross when the build fails.

Optionally you can add a build badge to the readme, which shows the state of the latest build.

About versioning: AppVeyor increases the build number for every commit and pull request for tracked branches, writing it as 4.0.{build} (with the current configuration). That number will eventually get pretty high. When you make a new release, and move on to 4.1 for example, you may want to reset this to 1 again.

### DIFF
--- a/Source/Dolphin-memory-engine.vcxproj
+++ b/Source/Dolphin-memory-engine.vcxproj
@@ -130,8 +130,8 @@
   <ItemGroup>
     <!-- All files except for GUICommon need to be Moc'ed -->
     <CustomBuild Include="GUI\**\*.h" Exclude="@(ClInclude)">
-      <Command>"$(QTDIR)bin\moc.exe" "%(FullPath)" -o ".\GeneratedFiles\$(Configuration)\moc_%(Filename).cpp" $(QtMocFlags) "-I." "-I$(QTDIR)include" "-I$(OutDir)GeneratedFiles\$(ConfigurationName)\." "-I$(QTDIR)include\QtCore" "-I$(QTDIR)include\QtGui" "-I$(QTDIR)include\QtWidgets"</Command>
       <Message>Moc'ing %(Filename)%(Extension)...</Message>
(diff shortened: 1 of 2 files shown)


aldelaro5:
Not sure about the fact it would be Windows only, ideally, one that supports both platform would be fine.

I was actually going to check into these CI, but I guess I just forgot.

If there's one that would support both Linux and Windows, that would be ideal.

Broxzier:
Idealy yes, but I haven't found one that does. Alternatives from the [GitHub market place](https://github.com/marketplace/category/continuous-integration) are:

- **TravisCI**: Linux/iOS, can perform tests
- **AppVeyor**: Windows, can perform test and deploy builds
- **CircleCI**: Linux/iOS, can perform tests
- **Buddy**: Windows, Seems to be limited with 120mins/month, not sure how fast we'll get over this
- **Semaphore**: Linux
- **Percy** and **Cloud 66 Skycap** are not free to use for Open Source projects.

Multiple CIs can be integrated together. Personally I lean towards adding Travis for Linux builds, just because I have good experiences with it. The others I have not worked with before.

aldelaro5:
ok, I will check them and see what I can do

PR #15, opened 2018-02-01
Make VS pick its default toolset

Broxzier:
As the title suggests, this PR makes any version of Visual Studio pick its default toolset, by setting it to `$(DefaultPlatformToolset)` rather than `v141`. This will also prevent conflicts later on when a new toolset gets released.

In the readme it's said that VS2015 users need to change some settings. If the platform toolset is the only change you meant, I'll update the readme too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,17 +22,14 @@ On Linux, installation of the Qt 5 package(s) is required.
 
 ## How to Build
 ### Microsoft Windows
-> *An installation of Microsoft Visual Studio 2015 or 2017 with Visual C++ is needed. Previous versions are untested. Building works best with VS2017, but small adjustments can be made for VS2015 to work.*
+This repository provides a solution file for Visual Studio 2015 and later. The Windows SDK Version 10.0.16299.0 (Windows 10 Fall Creators Update) is required*.
 
 Before proceeding, ensure the Qt submodule is initialized by running `git submodule update --init` at the repository's root. The files should appear at the `Externals\Qt` directory.
 
-Once complete, open Visual Studio then open the solution located in the `Source` directory.
+Once complete, open the solution file `Dolphin-memory-engine.sln` located in the `Source` directory with Visual Studio. Select the build configuration and build it.
 
-##### Visual Studio 2017
-Select the build configuration and build the solution.
-
-##### Visual Studio 2015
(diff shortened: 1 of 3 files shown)


aldelaro5:
it seems this does make it work, however, is it possible to have this done for the Windows SDK too?

Because it seems you still need to change it I was wondering if you could amend your commit with that.

Oh and yes I would like the readme change to be done, you are free to do it included with this PR or not.

Broxzier:
I've just searched through all the target and property sheets that are installed with MSVC, but all I could find is that the Windows SDK version defaults to 8.1 (stored in the `$(DefaultWindowsSDKVersion)` macro and even hardcoded as value in `Microsoft.Cpp.AppContainerApplication.props` and `Microsoft.Cpp.Common.props`), which unconveniently is not installed by default...

Then shall I update the readme to state the Windows 10.0.16299.0 SDK as a requirement, with a note that others may be used but not commited?

aldelaro5:
Yes please do so, sorry for taking so long to answer that one.  You don't really have to though, I could do that myself after I merge this PR (I need to do other docs update anyway).  Your choice.

README.md
```diff
@@ -19,11 +19,15 @@ On Linux, you need to install the Qt 5 package(s) for your respective distributi
 
 ## How to Build
 ### Microsoft Windows
-> *You will need Microsoft Visual Studio 2015 or 2017 with Visual C++ installed. Previous versions may work but are untested. The solution works best with VS2017, but with a small adjustment, it is possible to have VS2015 work.*
+#### Requirements
+- Visual Studio 2017
```
aldelaro5: It actually works on vs2015 too and it was tested too, just indicate it, otherwise LGTM.

aldelaro5:
Well, some other modifications of the readme got merged meaning you also have to rebase along with my comment.

Broxzier:
Dolphin requires Windows 10 SDK 10.0.15063.0 instead of 10.0.16299.0. Maybe it's nicer to switch to the same version?

aldelaro5:
This project doesn't have much link to Dolphin as far as building and developping goes, I just picked the version I happened to have, but the version doesn't seem that much of an importance, especially between these 2.  I wouldn't let the fact that dolphin uses another version be a concern unless it would have been officially supported by dolphin which it will not be.

Plus it seems the current version is more recent anyway so I think the current one is fine and changing it to the dolphin one won't do much.

Broxzier:
Sorry for taking so long to respond. I'll update the readme now. The 10.0.16299.0 Windows SDK is probably better anyway, since it comes with the Desktop C++ package from Visual Studio.

PR #16, opened 2018-02-03
Fix remaining memory leaks

Broxzier:
This fixes the following issues:
- The instance and the ram buffer from the `DolpingAccessor` class were not freed on exit. It's an annoying thing that comes with static singletones. I'd suggest refactoring this to something smarter or simpler in the future.
- Qt makes use of the composite design pattern, which means it doesn't require you to set parents or delete widgets manually. The parent/child relation is set when widgets are added to layouts or when setting one as the central widget. I've removed any manual setting of parents for those calls.

As additional change, I've renamed `main_layout` and `main_widget` to not use camelCase, like the rest of the document.

### DIFF
--- a/Source/DolphinProcess/DolphinAccessor.cpp
+++ b/Source/DolphinProcess/DolphinAccessor.cpp
@@ -28,6 +28,12 @@ void DolphinAccessor::init()
   }
 }
 
+void DolphinAccessor::free()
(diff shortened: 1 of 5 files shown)


Broxzier:
It happens :smile: Thanks for reviewing them all.

PR #21, opened 2018-02-13
Add support for localizing strings

Broxzier:
This mainly wraps string literals with `tr`, and for global string literals `QCoreApplication::Translate`. This also changes the strings themselves to take arguments using the QString::arg function, instead of appending them manually.

In `MemViewer::renderColumnsHeaderText` you'll see that the calculation has been altered, which has been done to take the spacing in front of the string into account without including it in the string itself.


aldelaro5:
Took the liberty to resolve the remaining conflicts, idk you could do it with the web editor.

Broxzier:
Oh, what was the conflict?

PR #22, opened 2018-02-19
Correct assigning window handle

Broxzier:
~~And add a comment for this function linking to the stackoverflow answer from where most of this code came.~~

It assigned to the same variable between two sequence points, which has undefined behaviour.


aldelaro5:
I did not took the code from SO, I actually read this from the msdn examples they provided and adapted them to my needs.

So I would not put that comment, it's not intricate enough in my opinion to really put the refference tbh, just the msdn docs was enough.

PR #36, opened 2019-01-12
 Bugfix: 'Add selection' adds wrong addresses

dreamsyntax:
Fixed minor oversight with addSelectedResultsToWatchList:
It was adding index 0 to num of selected regardless of what items were selected.

### DIFF
--- a/Source/GUI/MainWindow.cpp
+++ b/Source/GUI/MainWindow.cpp
@@ -157,7 +157,7 @@ void MainWindow::addSelectedResultsToWatchList(Common::MemType type, size_t leng
   QModelIndexList selection = m_scanner->getSelectedResults();
   for (int i = 0; i < selection.count(); i++)
   {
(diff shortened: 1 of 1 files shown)


aldelaro5:
I'll work more extensively in this project in about a month so I can't review this yet unfortunately (I am rusty and it requires some kind of testing), but for now, pls split both the bugfix and the sdk update.  I actually need to test more the sdk change anyway, it is true it has been a while the last version was released (almost a year actually).

I'll make sure to review it when I get to this project though!

dreamsyntax:
Done.

Unrelated:
I've been trying to reproduce the crash on delete that seems sporadic (memory leak somewhere perhaps?).
As well as the crash on enabling "this is a pointer".

When I catch said crash it appears to be attempting to free memory that has not been allocated.
I'll look more into that and hopefully have a PR up soon when I've fixed it.

PR #44, opened 2019-12-12
feature: add Copy+Paste at current index

dreamsyntax:
This is a feature my group has been wanting for a while. Finally got around to implementing it.
Super helpful when dealing with large DMWs as a group and cleaning up organization within DME.

Before:
- Using Ctrl+V would paste copied entry(ies) to the bottom of the DMW, no matter what
- Right-clicking a group or empty space allowed for "Paste" action via ContextMenu. Either adding to end of group or end of DMW, respectively.

After:
- Supports pasting entries to one element below currently selected item
- Enabled ContextMenu->Paste-ing on any element, following same behavior as keyboard 
- If a group is selected, append to end of group (just like ContextMenu->Paste behavior prior to PR)
- If no element is selected place at end of DMW

Misc:
- Removed 'canPasteInto' bool
- Modified signature of pasteWatchFromClipBoard to include an int (row)
- If null (empty list) assumes rootNode is safe and places element at end of rootNode
- Additional counter 'numberIterated' used to copy/paste collections of content in same order

Video demo of new behavior if my descriptions are not clear:
https://streamable.com/aroqh

### DIFF
--- a/Source/GUI/MemWatcher/MemWatchWidget.cpp
+++ b/Source/GUI/MemWatcher/MemWatchWidget.cpp
@@ -85,8 +85,11 @@ void MemWatchWidget::initialiseWidgets()
 
   QShortcut* pasteWatchShortcut =
       new QShortcut(QKeySequence(Qt::Modifier::CTRL + Qt::Key::Key_V), m_watchView);
-  connect(pasteWatchShortcut, &QShortcut::activated, this,
-          [=] { pasteWatchFromClipBoard(m_watchModel->getRootNode()); });
+  connect(pasteWatchShortcut, &QShortcut::activated, this, [=] {
+    pasteWatchFromClipBoard(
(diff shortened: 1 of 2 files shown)


aldelaro5:
I will try to review this more deeply this month (currently busy with reversing projects at the moment). From high level though, this seems fine to me.

aldelaro5:
Update: ik this repos is kinda dead rn and it's been so long I touched this project, it will probably take a while before I can probably review this. If I port this to dolphin, I will consider it.

aldelaro5 (approved):
There was a thing the formater found, but I realised this pr is too simple to pass up so I just tested it and it works fine so I am fine to merge it and I will myself pass the formater.

VERY SORRY it took SO LONG for me to check, but I just was fixing breaking issues so I am reviewing prs that have been pending for a while

Issue #35, opened 2018-10-26
"Cannot hook to Dolphin, the process is running, but no emulation has been started" error


calvinpelletier:
I'm getting "Cannot hook to Dolphin, the process is running, but no emulation has been started" even though an emulation is running. Any thoughts? I'm using Dolphin 5.0.

aldelaro5:
I just want to comment on this saying you are not the only person that reported issues with the base address since. The tool I linked above is a good one to debug this because you can check which pages looks similar, the issue is more that it's complicated to use for end users so I will be checking if I should simply log the pages mapped by Dolphin and display them for debugging purposes so that way, I can see why no pages were chosen for the heuristics and Dolphin can report the TRUE address using the logger and the instructions above.

But yeah, so far, this is a mystery.

moomath:
I am running this on linux and it took a while to set up but I was wondering if any information has been released on what to do? Is there just no solution yet? Thank you.

aldelaro5:
Still unsure how to replicate this unfortunately. It seems to be very inconsistent across users.

aldelaro5:
@moomath oh, have you tried this from the readme?

Troubleshouting

On Linux, the program may require additional kernel permissions to be able to read and write memory to external processes (which is required to read and write the memory of Dolphin). If nothing happens to Dolphin, but the program frequently unhooks itself, the program is missing the required permissions. Grant these permissions by running the following command as root:

setcap cap_sys_ptrace=eip DME

Where DME is the path of the Dolphin Memory Engine executable. This should fix the permission problem for future executions.

moomath:
@aldelaro5 yes, i was able to get it to work with your help, haha, since we were talking on discord just now.

csunday95:
@aldelaro5 I am experiencing the same "the process is running, but no emulation has been started" issue despite a GC game being emulated. My setup is

Dolphin 5.0 Stable

Ubuntu 18.04.3 LTS

Dolphin-memory-engine Beta 0.5 (built from source)

I have performed the setcap command outlined above, but it had no effect.

Gamebuster19901:
I'm also experiencing this issue, where it says "the process is running, but no emulation has been started" despite a wii game being emulated.

My setup is:

Dolphin 5.0 Stable

Ubuntu 18.04.4 LTS

Dolphin-memory-engine Beta 0.5.

I have also executed the setcap command above as the root user with no effect.

I have also tried running memory engine as sudo with no effect.

[...]

PR #52, opened 2020-11-17
Fix MEM2 Wii only RAM not found even when playing Wii games

### LINKED ISSUE #46: MEM2 Wii only RAM not found even when playing Wii games
I think it may be due to Dolphin's updates, but as of a few weeks ago it stopped being able to view the Wii only RAM at +0x90000000. It always says that the memory region is absent.

NicholasMoser:
# Fixes #46 

## Background

As per #46, we know that https://github.com/dolphin-emu/dolphin/pull/8537 broke Dolphin Memory Engine's ability to read MEM2. When I debug Dolphin Memory Engine, I see that it's not entering this if check:

https://github.com/aldelaro5/Dolphin-memory-engine/blob/0b2aea13fb5c12363376247e50dc326383bf40c1/Source/DolphinProcess/Windows/WindowsDolphinProcess.cpp#L59

The above if check looks `0x10000000` bytes away from the start of MEM1 for MEM2. So I presume that MEM2 is no longer `0x10000000` bytes away from the start of MEM1. I know that MEM2 has a size of `0x04000000`, so I used that to find if there's a new offset. Below I have a picture of MEM1 and MEM2 on dolphin versions prior to https://github.com/dolphin-emu/dolphin/pull/8537:

image

If you subtract the MEM1 BaseAddress from the MEM2 BaseAddress, you'll indeed get a distance of `0x10000000`. So now let's see what the distance is for newer versions of Dolphin:

image

It is now a distance of `0x2000000` bytes, so we need to change the code to reflect that.

## The Fix

First, change the MEM1 start to MEM2 distance from `0x10000000` to `0x2000000`. This includes changing `Common::dolphinAddrToOffset(Common::MEM2_START)` since that returns `0x10000000`.

We also need to add a check for the region size of MEM2 (`0x04000000`). This is because for Wii games, the MEM2 is now directly after MEM1. But for GameCube games this is not the case. Instead, some other region is with a size of `0x2000000`. We simply ignore that using the region size check.

## Testing

GameCube game working as expected:
image

Wii game working with MEM2:
image


NicholasMoser:
Note: I only fixed Windows in this PR. I don't even know if Linux is also broken. If it is, I presume the fix will be the same since I also see `LinuxDolphinProcess.cpp` using `0x10000000` to find MEM2.

bkacjios:
Thanks for this! Just figured I'd let you know that if you do a scan with mem2 available, it seems to fail and unhook itself.

> 
> 
> Note: I only fixed Windows in this PR. I don't even know if Linux is also broken. If it is, I presume the fix will be the same since I also see `LinuxDolphinProcess.cpp` using `0x10000000` to find MEM2.

Figured I'd let you know Linux isn't broken. Been using it since Windows hasn't been working.

NicholasMoser:
Thanks for the information. I'll take a look within the next week or so and try and fix that issue. I probably should also do some more extensive testing...

NicholasMoser:
Few notes:
1. I'm realizing that my original proposal likely would break Linux, so I'll make sure that any Windows specific logic is behind an `#ifdef _WIN32`.
2. It looks like in some cases MEM2 may actually be **before** MEM1 in memory. Not sure how dolphin determines this, but I'll make sure I account for it (occurred with Dolphin 5.0-13022 on Windows)

NicholasMoser:
@bkacjios
I'm still in the process of testing my changes, but I believe I have it working now for both Wii and GameCube. Feel free to try it.

Basically I had to detect the location of MEM2, calculate the distance between MEM1 and MEM2, and use this for all calculations involving MEM2.

bkacjios:
Scanning looks all good now! If you want, I can check to make sure Linux still works.

NicholasMoser:
I don't expect it to, since the LinuxDolphinProcess doesn't calculate the MEM2 distance. I'll work on getting that working at some point ;)

JimKatz:
Do you plan to release a build of this? I would love to try it!

NicholasMoser:
@JimKatz I can't officially release it since the repository doesn't appear to be active anymore. Let me try reaching out to @aldelaro5 and seeing if he's willing to add anyone as an owner to the repo.

I suppose I could release it under my fork at least? Or just provide binaries on this pull request to test out.

aldelaro5:
Honestly....I really don't know what to do.

The reality is that as of now...yeah the repo is kinda dead and unfortunately, I don't plan to come back to it in the near future. I was planning to attempt to merge this within dolphin (which would be ideal since maintenance won't be an issue if it's done), but currently, there's just no way I can maintain this repo which is sad because there's A LOT more people that uses it than I expected :(

I am tempted to say release a fork because the reality is even if I wanted to merge, getting back into this project would require me to remind myself how it works since...it's been a WHILE I used it.

I am unsure about adding a owner on this one (since I care a bit more about this one), but I will say this: I am open you fork or even do the dolphin merging work because there's always a chance I pick this repo later in the future, but as of now....there's just no way I can maintain it.

NicholasMoser:
Thanks for the response.

I'd be willing to look into seeing what it would take to merge this into Dolphin. The code could even be cleaner since it has direct access to MEM1 and MEM2. I probably won't start at least for a few months though.

aldelaro5:
from what I heard on their irc, it seems possible perf wise, but you might have to deal with this being an opt-in (cause it WILL tank the perf if you allow to read dozens of values each timer update).

The reason in the first place I made this project is CE was just frustrating to use with dolphin and even today, it has very lacking memory search/watch/scanning features (it does have them, but they suck like they don't even work in realtime). I realised quickly this project is just....not going to go far until it's inside dolphin hence why I had plans to do it, but long story short, I played bug fables and I'm not fully focussed on that game as well as other stuff (like learning unity).

it's like I want to put more love into DME, but can't :(

NicholasMoser:
Totally understand, thanks for the time you've put into it. It's been a huge help to the modding communities I'm a part of :)

.gitignore
```diff
@@ -14,3 +14,4 @@ Source/*.VC.opendb
 Source/*.VC.db
 Source/.vs/
 Source/.vscode/
+/.vs/
```
aldelaro5, line 17: I think this is not necessary, I usually always open the source folder and not the entire thing

aldelaro5 (changes requested):
This is a conplex pr to review so I might do other reviews as I do my test and start to deeply delve into it, but you can fix this one as a start.

aldelaro5:
> I don't expect it to, since the LinuxDolphinProcess currently doesn't calculate the MEM2 distance. I'll work on getting that working at some point ;)

I will test on linux on my end, it indeed doesn't work, but it was already existing so it's fine so for now, I will exclusively review this pr on Windows

aldelaro5 (changes requested):
OK so add this review to my previous comment, sorry for splitting, but I am being overwhelmed checking this project after so long :)

Also, pls run clang-format on your file (there is a bash file in the tools folder and Visual Studio has an extension you can use to format the files). Here are the files where you need to run clang-format after you have addressed my review:

- DolphinAccessor.cpp
- MemoryScanner.cpp
- MemViewer.cpp
- MemWatchEntry.cpp
- WindowsDolphinProcess.cpp

Other than that, it's all LGTM. I actually really like your distance-agnostic approach because it doesn't break compatibility with older versions which is what I was worried with.

Once this is all addressed, I am ready to merge. Since this is a big fix, I ask that you do this as soon as possible.

aldelaro5:
Ok I have done the Linux works necessary to fix it and to make it MEM2 distance agnostic when you merge the pr, I can't push yet until this is merged. It also fixes #35 too so basically, I REALLY need this merged for both of these bugs to be fixed.

Btw, I found out that they map MEM1 twice due to fastmem and it seems MEM2 can be mapped three times. I also did got into the situation where MEM2 could be before MEM1 so yeah they definitely did some interesting changes.

NicholasMoser:
Thanks for the review.
I'm going to be pretty busy over the next handful of days, but likely can get to this in a week or so. If you need this sooner please feel free to push to this branch with the necessary changes :)

.gitignore
NicholasMoser: Fixed in https://github.com/aldelaro5/Dolphin-memory-engine/pull/52/commits/f2d8ea6314fed8dfc59a79e94d281f9069ba69a2 and https://github.com/aldelaro5/Dolphin-memory-engine/pull/52/commits/d8da0b56dce8132f61f741511862868bf0cf4b06

Source/DolphinProcess/Windows/WindowsDolphinProcess.cpp
```diff
@@ -48,35 +48,34 @@ bool WindowsDolphinProcess::obtainEmuRAMInformations()
 {
   MEMORY_BASIC_INFORMATION info;
   bool MEM1Found = false;
+  bool MEM2Found = false;
```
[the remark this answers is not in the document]
NicholasMoser: Fixed in https://github.com/aldelaro5/Dolphin-memory-engine/pull/52/commits/218f4ac4b094da53cbc0c686e5f4e679082b4678

NicholasMoser:
I've made the requested changes. Please let me know if any further changes are necessary.

aldelaro5 (approved):
LGTM, since this has already been tested, merging now.

PR #60, opened 2022-06-16
Bouquet of UI enhancements

cristian64:
A number improvements to the UI:

- Add splitter in main window to allow resizing:

- Store window state and window geometry in settings file. State is restored on startup.
- Provide a window title to all dialogs (**Add Watch**, **Edit Watch**, **Memory Viewer**, **Dolphin Memory Engine**, ...).
- Rework **Add Watch** dialog with a proper form layout:

image

- Memory watch list is now saved as part of the settings; list is restored on startup.
- Support for memory addresses that start with the `0x` prefix.
- Rows in the memory watch table are no longer sortable: users could mistakenly reordered their curated list.

Bug fixes:

- Crash involving a double deallocation in `DlgAddWatchEntry`.

### DIFF
--- a/Source/GUI/MainWindow.cpp
+++ b/Source/GUI/MainWindow.cpp
@@ -23,6 +23,11 @@ MainWindow::MainWindow()
   DolphinComm::DolphinAccessor::init();
   makeMemViewer();
   firstHookAttempt();
+
+  if (SConfig::getInstance().getMainWindowGeometry().size())
+    restoreGeometry(SConfig::getInstance().getMainWindowGeometry());
+  if (SConfig::getInstance().getMainWindowState().size())
+    restoreState(SConfig::getInstance().getMainWindowState());
 }
 
 MainWindow::~MainWindow()
@@ -399,6 +404,8 @@ void MainWindow::closeEvent(QCloseEvent* event)
 {
   if (m_watcher->warnIfUnsavedChanges())
   {
+    SConfig::getInstance().setMainWindowGeometry(saveGeometry());
+    SConfig::getInstance().setMainWindowState(saveState());
     m_viewer->close();
     event->accept();
   }
--- a/Source/GUI/MainWindow.cpp
+++ b/Source/GUI/MainWindow.cpp
@@ -17,6 +17,7 @@
 
 MainWindow::MainWindow()
 {
+  setWindowTitle("Dolphin Memory Engine");
   initialiseWidgets();
   makeLayouts();
   makeMenus();
@@ -372,7 +373,7 @@ void MainWindow::onOpenSettings()
 
 void MainWindow::onAbout()
 {
-  QString title = tr("About Dolphin memory engine");
+  QString title = tr("About Dolphin Memory Engine");
   QString text =
       "Beta version 0.6.0\n\n" +
       tr("A RAM search made to facilitate research and reverse engineering of GameCube and Wii "
--- a/Source/GUI/MainWindow.cpp
+++ b/Source/GUI/MainWindow.cpp
@@ -4,6 +4,7 @@
 #include <QMenuBar>
 #include <QMessageBox>
 #include <QShortcut>
+#include <QSplitter>
 #include <QString>
 #include <QVBoxLayout>
 #include <string>
@@ -134,16 +135,24 @@ void MainWindow::makeLayouts()
   QFrame* separatorline = new QFrame();
   separatorline->setFrameShape(QFrame::HLine);
 
+  QSplitter* splitter = new QSplitter(Qt::Vertical);
+  splitter->addWidget(m_scanner);
+  splitter->addWidget(m_watcher);
+
+  if (SConfig::getInstance().getSplitterState().size())
+    splitter->restoreState(SConfig::getInstance().getSplitterState());
+
+  connect(splitter, &QSplitter::splitterMoved,
+          [splitter = splitter]()
+          { SConfig::getInstance().setSplitterState(splitter->saveState()); });
+
   QVBoxLayout* mainLayout = new QVBoxLayout;
   mainLayout->addWidget(m_lblDolphinStatus);
   mainLayout->addLayout(dolphinHookButtons_layout);
+  mainLayout->addWidget(m_btnOpenMemViewer);
   mainLayout->addWidget(m_lblMem2Status);
   mainLayout->addWidget(separatorline);
-  mainLayout->addWidget(m_scanner);
-  mainLayout->addSpacing(5);
-  mainLayout->addWidget(m_btnOpenMemViewer);
-  mainLayout->addSpacing(5);
-  mainLayout->addWidget(m_watcher);
+  mainLayout->addWidget(splitter);
 
   QWidget* mainWidget = new QWidget();
   mainWidget->setLayout(mainLayout);
--- a/Source/GUI/MainWindow.cpp
+++ b/Source/GUI/MainWindow.cpp
@@ -29,6 +29,8 @@ MainWindow::MainWindow()
     restoreGeometry(SConfig::getInstance().getMainWindowGeometry());
   if (SConfig::getInstance().getMainWindowState().size())
     restoreState(SConfig::getInstance().getMainWindowState());
+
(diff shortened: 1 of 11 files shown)


cristian64:
I thought I'd open this PR with a few UI improvements that I wish had been available when I was using this tool.

I'm not expecting much traction, though, so I will abandon the PR after a few days, as I wouldn't like to increase the number of stale PRs.

cristian64:
As promised, no need to contribute to the list of moldering PRs. **Abandoned**.

dreamsyntax:
Reopening as I will be revisiting this / using as a reference.

dreamsyntax:
@cristian64 Would you be willing to rebase this PR? Looked it over, I'm basically fine merging as is once resolved. If you have no interest in this project anymore I can just cherry-pick your commits onto it.

cristian64:
Sounds good to me. Give me a couple of days to rebase.

Source/GUI/MainWindow.cpp
```diff
@@ -15,12 +16,22 @@
...
   firstHookAttempt();
+
+  if (SConfig::getInstance().getMainWindowGeometry().size())
+    restoreGeometry(SConfig::getInstance().getMainWindowGeometry());
+  if (SConfig::getInstance().getMainWindowState().size())
+    restoreState(SConfig::getInstance().getMainWindowState());
+
+  m_watcher->restoreWatchModel(SConfig::getInstance().getWatchModel());
+
+  hookIfNotHooked();
```
dreamsyntax, line 34: A few lines above this the firstHookAttempt is already called. Is there a particular reason we should keep this?

dreamsyntax (changes requested):
Sounds good, thank you.

Source/GUI/MainWindow.cpp
cristian64: At the moment, `hookIfNotHooked()` schedules _itself_ for another run in 3000 ms, so it needs this first call to start the loop.

This call could be changed to
```cpp
QTimer::singleShot(3000, this, &MainWindow::hookIfNotHooked);
```
, so the loop is started without trying to hook.

I think my preference would be to remove the `firstHookAttempt()` function, and add the `updateMem2Status()` call directly in `hookIfNotHooked()`, and then we can keep this call to `hookIfNotHooked()` here, which then starts the cycle.

dreamsyntax:
If it makes sense to add it here, the ability to disable the watchlist sort would be appreciated. 

Otherwise Ill just open a PR after this is merged.

cristian64:
> If it makes sense to add it here, the ability to disable the watchlist sort would be appreciated. (Clicking the tab column names)
> 
> Otherwise Ill just open a PR after this is merged.

I added a commit with the `setSortingEnabled()` line removed. (I agree it was a bit risky; the user could easily lose the order.)

Source/GUI/MainWindow.cpp
cristian64: This has been done.

(I finally remembered why `updateMem2Status()` was not being called in `hookIfNotHooked()`: it's already called in `onHookAttempt()`.)

cristian64:
It should have been rebased now.

Really glad the project has been resuscitated; I still use it almost every day. There are other improvements that I'd like to make, but I guess it can be done in follow-up PRs. Some of these changes are:

- Reduce padding/margins in the UI to fit more useful data. It feels like Qt wastes much space at the moment.
- Remove some of the churn in the GUI, like those big buttons, **Hook** or **Open memory viewer**, that could be moved into the top-bar menu, specially now with the auto-hook mechanism.
- Fix some crashes that occur in certain conditions when Dolphin is closed while DME is hooked.

dreamsyntax:
> It should have been rebased now.
> 
> Really glad the project has been resuscitated; I still use it almost every day. There are other improvements that I'd like to make, but I guess it can be done in follow-up PRs. Some of these changes are:
> 
> - Reduce padding/margins in the UI to fit more useful data. It feels like Qt wastes much space at the moment.
> - Remove some of the churn in the GUI, like those big buttons, **Hook** or **Open memory viewer**, that could be moved into the top-bar menu, specially now with the auto-hook mechanism.
> - Fix some crashes that occur in certain conditions when Dolphin is closed while DME is hooked.

Glad to hear it. I welcome changes, but please hold off Until Qt6 merge (soon). You can base your PR off Qt6 PR if you dont want to wait.

Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
```diff
@@ -33,10 +37,11 @@ void DlgAddWatchEntry::initialiseWidgets()
   connect(m_chkBoundToPointer, &QCheckBox::stateChanged, this,
           &DlgAddWatchEntry::onIsPointerChanged);
 
-  m_lblValuePreview = new QLabel("", this);
+  m_lblValuePreview = new QLineEdit("", this);
+  m_lblValuePreview->setReadOnly(true);
 
   m_txbAddress = new QLineEdit(this);
-  m_txbAddress->setMaxLength(8);
+  m_txbAddress->setMaxLength(10);
```
dreamsyntax, line 44: I'm not following why this is 10 now instead of 8? Wouldnt the address always be XXXXYYYY ?

Source/GUI/MainWindow.cpp
```diff
@@ -159,12 +163,15 @@ void MainWindow::makeMemViewer()
...
-  onHookAttempt();
-  if (DolphinComm::DolphinAccessor::getStatus() ==
+    QTimer::singleShot(3000, this, &MainWindow::hookIfNotHooked);
+
+    if (DolphinComm::DolphinAccessor::getStatus() !=
       DolphinComm::DolphinAccessor::DolphinStatus::hooked)
-    updateMem2Status();
+    {
+      onHookAttempt();
+    }
```
dreamsyntax, line 174: Looking at this again, I have some concerns about constantly making a QTimer singleShot. I'll have to profile and test in real world use.

Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
cristian64: It's mentioned in the commit: to allow for the `0x` prefix.

Source/GUI/MainWindow.cpp
dreamsyntax: After testing this build in real world use, I absolutely want there to be a setting to **Turn Off** auto hooking.
In my own use case I regularly purposely unhook with multiple DMEs up with different freeze values already pre-loaded.

dreamsyntax (changes requested):

Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
dreamsyntax: Thanks for clarifying.

Source/GUI/MainWindow.cpp
cristian64: Performance wise it's not going to be a concern; it wouldn't be different to a `QTimer`. I wouldn't disagree that a `QTimer` may be a more orthodox approach, though.

cristian64: I think it's a sensible request. I was hoping to add options for disabling auto-hooking once the big **Hook** button is moved to a different place, but I can do it now (or take this out of this PR).

dreamsyntax: It depends if you think its easy enough to implement/add.

I would suggest doing something like this (adding to the settings.ini we generate)
https://github.com/aldelaro5/Dolphin-memory-engine/pull/72/files

cristian64: I'll take auto-hooking out, and open a different PR for it.

cristian64: Commit removed from branch. I will open a different PR shortly (which can then stay on hold until the Qt 6 work is completed).

dreamsyntax (approved):

PR #61, opened 2022-06-21
Features: CopyMemoryRange (new View); Search Range; Undo Search

roeming:
Added a dialog which can be found under the View menu for copying segments of memory into formatted byte strings.
These byte string options are:

- Byte String
- Byte String (No Spaces)
- Python Byte String
- Python Byte List
- C Array

The dialog takes a base address, a count (hex or decimal), and a method for formatting the bytes. Upon clicking 'Apply', the bytes will be shown as a formatted string in the output box. The output only updates when 'Apply' is clicked. If the byte format changes, the old data will be reformatted to the new format.

The boxes are error checked for valid inputs. I created my own error checking functions for valid hex/decimal strings and for hex string->u32 and decimal string -> u32 because I couldn't find a common function for doing so.

This was my first time working with Qt, so let me know if there's a more proper way to work with some of these functions and classes.

I was unsure of the "Moc'ing" that was happening during the build, so I'm not sure if I'm building in a way that conforms to that style. So that may be another thing worth looking at.

### DIFF
--- a/Source/Dolphin-memory-engine.vcxproj
+++ b/Source/Dolphin-memory-engine.vcxproj
@@ -118,6 +118,7 @@
     <ClInclude Include="**\*.h" Exclude="GUI\**\*.h" />
     <!-- These files don't need to be Moc'ed -->
     <ClInclude Include="GUI\GUICommon.h" />
+    <ClInclude Include="GUI\MemCopy\DlgCopy.h" />
     <ClInclude Include="GUI\MemWatcher\Dialogs\DlgAddWatchEntry.h" />
     <ClInclude Include="GUI\MemWatcher\MemWatchDelegate.h" />
     <ClInclude Include="GUI\MemWatcher\MemWatchTreeNode.h" />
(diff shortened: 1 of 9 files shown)


dreamsyntax:
@roeming It seems you also added Undo and Scanner Range feature in this PR. If you are willing to rebase, we can use this after I do some tests with it.

roeming:
Rebased

dreamsyntax:
@roeming failing linux build:
```
...Dolphin-memory-engine/Source/GUI/MemCopy/DlgCopy.h:7:10: fatal error: ..\GUICommon.h: No such file or directory
```

dreamsyntax:
In general, the features work well and the code is fine.
Before merging this however, could you please change the window type to not include the (?) button. It doesn't seem to function anyhow.

image

dreamsyntax:
> In general, the features work well and the code is fine. Before merging this however, could you please change the window type to not include the (?) button. It doesn't seem to function anyhow.
> 
> image

Actually, you can disregard this. It seems to be something that is automatically removed once we move to Qt 6.
The other windows have the same (?) button.
image

dreamsyntax:
I'm just going to merge this and fix it since I saw someone else opened a PR on your fork with the CMake change:
https://github.com/roeming/Dolphin-memory-engine/pull/1/commits/a62feb5bd7c0379027fd4497ff71e4ec0ef13652

Thanks again for your effort.

dreamsyntax (approved):

PR #65, opened 2023-05-25
Added lock/unlock for selecting multiple items

MarkAlarm:


dreamsyntax:
+1. Intended to do this myself a long time ago.

dreamsyntax (approved):
LGTM; Ideally pressing Lock with multi-selection on any of the elements would lock. But seeing as we have no context menu option for Lock I'm fine with this as is.

Issue #80, opened 2023-11-04
Feature: Portable Scope


dreamsyntax:
* With the release of 0.7.0 settings.ini is now stored in user or system appdata equivalent.
* When portable.txt is in the directory, instead save any configuration to the current folder.
* Match Dolphin behavior with portable.txt

PR #81, opened 2023-11-04
AppImage Build

qurious-pixel:
Can be used to provide portability 
Issue: #86


.github/assets/Dolphin-memory-engine.desktop
```diff
@@ -0,0 +1,8 @@
+[Desktop Entry]
+Type=Application
+Name=Dolphin memory engine
```
dreamsyntax, line 3: Dolphin Memory Engine for consistency

.github/assets/Dolphin-memory-engine.png
dreamsyntax, line 1: I'll be re-doing this screenshot, though I suppose I can just update this later.

Is there a reason it has to be 512x512 for app image?

.github/assets/appimage.sh
```diff
@@ -0,0 +1,22 @@
...
+chmod +x AppDir/AppRun
+
+chmod a+x appimagetool linuxdeploy linuxdeployqt
+ARCH=x86_64 ./linuxdeploy --appdir=AppDir
+
+cp .github/assets/"${BINNAME}".png AppDir/
+cp .github/assets/"${BINNAME}".desktop AppDir/
+mkdir -p AppDir/usr/share/applications && cp ./AppDir/"${BINNAME}".desktop ./AppDir/usr/share/applications
+mkdir -p AppDir/usr/share/icons && cp ./AppDir/"${BINNAME}".png ./AppDir/usr/share/icons
+mkdir -p AppDir/usr/share/icons/hicolor/512x512/apps && cp ./AppDir/"${BINNAME}".png ./AppDir/usr/share/icons/hicolor/512x512/apps
```
dreamsyntax, line 19: With the new icon added in #84 , can we apply it to the app image? Does it change this?

dreamsyntax (changes requested):

dreamsyntax:
Lastly, regarding
> https://github.com/aldelaro5/Dolphin-memory-engine/issues/80 
This is not actually related to #80, that is to add an extra option for User / System scope.
I didn't have an issue open for AppImage at the time.

.github/assets/Dolphin-memory-engine.desktop
qurious-pixel: Upper is better. I was matching the binary.

.github/assets/Dolphin-memory-engine.png
qurious-pixel: Does not have to be 512x512. Chose for visibility. Will use the. ico/png from #84

.github/assets/Dolphin-memory-engine.desktop
dreamsyntax: I think I'll change the binary and repo to be all lower to match usual app naming and repo standards, but any displayed situation should be Dolphin Memory Engine I think, or DME. Whatever makes more sense.

dreamsyntax:
FYI, pipeline failing. Let me know if you dont have access
> Run .github/assets/appimage.sh
/home/runner/work/_temp/99cef56d-a49b-4256-a0cc-d88e96d18449.sh: line 1: .github/assets/appimage.sh: Permission denied
Error: Process completed with exit code 126.

qurious-pixel:
I didn't have the exec bit set on appimage.sh, and I had pushed the fix, but guess it didn't take it. I can set it again and push all as a single commit,  or individually with the requested changes.

dreamsyntax:
> I didn't have the exec bit set on appimage.sh, and I had pushed the fix, but guess it didn't take it. I can set it again and push all as a single commit, or individually with the requested changes.

Either is fine, though generally I prefer single commits.

dreamsyntax:
For future reference, you can just git rebase onto latest master to avoid combining multiple merges
`git fetch --prune`
`git checkout master`
`git pull`
`git checkout your-branch`
`git rebase master`
-> Resolve conflicts if there are any.

Now that it has a merge commit in it though, it may be hard to correct without redoing the changes or cherry-picking your specific commits. Sorry for the hassle, if you want I can just do this before merging.

qurious-pixel:
Yes please fix,  I am on mobile for now. I can rebase properly when I have a chance.

dreamsyntax:
Tested it, it looks like the icon does not work but everything else works.
You mentioned we should probably use ubuntu 20.04 as a builder. Does that increase compatibility with certain systems or something? If yes, I'm fine changing it.
image

qurious-pixel:
The lower libc requirement increases compatibility, but after rethinking, there are 6 months left on that LTS, it is easier to leave on the current build image (22.04).

Will look at the icon.

dreamsyntax (approved):
The icon still does not work, but it seems there's issues with Windows displaying the icon too. For now merging this

dreamsyntax:
Update, my mistake the icon is working with the final version merged.
Was my system cache.

Thanks for all your effort! @qurious-pixel 
image

cristian64:
> [...], there are 6 months left on that LTS, it is easier to leave on the current build image (22.04).

If we were talking about Ubuntu 20.04 LTS, official support ends in April 2025; that's 17 months from now. And then there is a period of security updates for another 5 years.

Unsure how easy it'd be to get Qt 6 on Ubuntu 20.04 in GitHub Actions, though.

qurious-pixel:
> If we were talking about Ubuntu 20.04 LTS, official support ends in April 2025; that's 17 months from now. And then there is a period of security updates for another 5 years.
> 
> Unsure how easy it'd be to get Qt 6 on Ubuntu 20.04 in GitHub Actions, though.

Yes, you are right, I had miscounted 4 years, instead of 5.
On the CI, I can get support for QT6 fairly easily. It will be more difficult with the package manager, because the backports are not often updated. Ubuntu 18.04 had an issue, because the precompiled used a later libc version that was incompatible, requiring 60GBs to compile the package.
Qt6 support, libc version, and old enough compiler (something available on the latest point release of the supported system) are usually the factors to consider.

cristian64:
If the official precompiled binaries from the Qt Online Installer are available for direct download (are they?), they should work with Ubuntu 20.04, which uses glibc 2.31; the precompiled binaries (GCC x64) in Qt 6.5.3 are compiled against glibc 2.28.

> On the CI, I can get support for QT6 fairly easily. [...]

Are there APT-based repositories that could be used in Ubuntu 20.04? Or do you have another source in mind?

qurious-pixel:
Ubuntu 18.04 was glibc 2.27, so it was not so easy to upgrade to Qt6.

Ubuntu Focal PPA:
```
sudo add-apt-repository ppa:rncbc/qt6.6-static-focal
sudo apt update
```


The CI could use a Github Action 
```
    - name: Install Qt
      uses: jurplel/install-qt-action@v3
      with:
        aqtversion: '==3.1.*'
        version: '6.6.0'
        host: 'linux'
        target: 'desktop'
        arch: 'gcc_64'
```

cristian64:
I was hoping we would go as far as Ubuntu 20.04. Are you aiming for Ubuntu 18.04?

PR #82, opened 2023-11-04
Add auto-hooking mechanism.

cristian64:
- The **Hook** and **Unhook** buttons have been moved to the new **Dolphin** top-bar menu.
- A new **Auto-hook** checkable menu action (checked by default) has been added to the menu too.

Dolphin Memory Engine - Auto-hooking Mechanism

Bonus:
- The **Open memory viewer** button has been moved to the existing **View** top-bar menu.
- The **Copy Memory Range** dialog is now created only once on startup, and its enable state is synced with hooked state. Its associated menu action in the **View** menu is now enabled/disabled too.

### DIFF
--- a/Source/GUI/MainWindow.cpp
+++ b/Source/GUI/MainWindow.cpp
@@ -63,6 +63,7 @@ void MainWindow::makeMenus()
   m_actHook = new QAction(tr("&Hook"), this);
   m_actUnhook = new QAction(tr("&Unhook"), this);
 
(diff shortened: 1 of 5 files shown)


dreamsyntax (approved):
I'm not exactly liking the removal of the single click flow for hook/unhook, but I also acknowledge for most users it's redundant.

I suppose this is fine for now, I think we should consider redesigning the general view anyway. The info about where the address is hooked to could maybe just be under an info dialog popup. Hook status should definitely still be displayed, but I suppose that's for another PR.

PR #91, opened 2023-11-05
Specify `Release` build type for Linux build in GitHub Action.

cristian64:
Now `-O3 -DNDEBUG` are passed in to the compiler.

Additionally, generator-agnostic commands are now used, as opposed to invoking `make` directly. Also, the build is now done in parallel.


dreamsyntax (approved):
Good catch, thanks.

PR #96, opened 2024-01-12
macOS Support

EndangeredNayla:


Source/DolphinProcess/Darwin/DarwinDolphinProcess.h
```diff
@@ -0,0 +1,26 @@
...
+#include <mach/mach.h>
+
+namespace DolphinComm
+{
+class DarwinDolphinProcess : public IDolphinProcess
+{
+public:
+  DarwinDolphinProcess()
+  {
+  }
```
dreamsyntax, line 15: minor nitpick, not a fan of this but if its doing this via autoformat / or other files are like this, then whatever

dreamsyntax (changes requested):
Minor change requested; Thank you very much for this PR.

Source/DolphinProcess/Darwin/DarwinDolphinProcess.h
EndangeredNayla: > other files are like this, then whatever

this is exactually the case

.github/workflows/build.yml
```diff
@@ -40,8 +41,7 @@ jobs:
 
       - name: Set Artifact Name
         id: set-artifact-name
-        run: echo "artifact_name=$(echo "windows-${{ matrix.configuration }}" | tr '[:upper:]' '[:lower:]')" >> $GITHUB_OUTPUT
-        shell: bash
+        run: echo "artifact_name=$(echo "windows-${{ matrix.configuration }}" | tr '[:upper:]' '[:lower:]')" >> $GITHUB_ENV
```
dreamsyntax, line 44: This breaks the build for Windows.

$GITHUB_OUTPUT is a specific var name 
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions

Please restore this, other than that LGTM. Ready to merge once all pipelines pass and I get a chance to test on all platforms.

dreamsyntax (changes requested):

EndangeredNayla:
Okay should be all set. For whatever reason on macos it always detects the extended Wii memory but it's only a visual thing and the program still works.

EndangeredNayla:
I attached a debug dolphin binary to the readme for people who don't know how to compile a debug copy of dolphin on macos.

EndangeredNayla:
Codesigning should not be nessary for this app to work so tell me if u have an issue, I just wonder if my global certificate may be spoofing that so I would like some other Mac user to test it first

campital:
If I recall correctly, the macOS port (#50) required a removal of the cache mechanism for performance reasons. Not sure whether this is true for the newer versions, but I can test this soon. Also, @dreamsyntax I'm just seeing your comments from a few months ago on my PR. I can rebase it within the next few days or we can go with this PR.

dreamsyntax:
> If I recall correctly, the macOS port (#50) required a removal of the cache mechanism for performance reasons. Not sure whether this is true for the newer versions, but I can test this soon. Also, @dreamsyntax I'm just seeing your comments from a few months ago on my PR. I can rebase it within the next few days or we can go with this PR.

I would prefer to go with this as its actively being used. If we identify issues we can iterate using your PR as reference.

EndangeredNayla:
Also not sure where the code signing is coming from on the other PR as this one actively doesn't need those entitlements just dolphin itself does.

dreamsyntax:
@EndangeredNayla @campital please collaborate on the resolution
Are we going to change this PR to merge into campital's PR (for the GH action) and any README adjustments?

dreamsyntax:
I'm not sure if discussion was had off GH, but seeing as this is now inactive and #50 is nearing completion, closing this.
Make a new PR with README changes and anything else if you wish based on final changes in #50.

PR #97, opened 2024-01-17
Remove primary cache

campital:
Removing the primary cache yields a very significant performance improvement on all platforms (90% reduction in memory usage, CPU usage is no longer 100%). The cache does not seem to serve any actual purpose. Memory viewer GUI is now much more usable on Windows, though there remains some sluggishness due to the high memory refresh rate.

Also fixed a crash bug in `DlgAddWatchEntry.cpp`, an unhooking bug that occurred when MEM2 memory watches were still present while loading a game without MEM2, and a bug that prevented DME from discovering Dolphin when closed and relaunched.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,5 +14,5 @@ Source/*.tlog
 Source/*.VC.opendb
 Source/*.VC.db
 Source/.vs/
-Source/.vscode/
+.vscode/
 .idea/*
--- a/.gitignore
+++ b/.gitignore
@@ -14,5 +14,6 @@ Source/*.tlog
 Source/*.VC.opendb
 Source/*.VC.db
 Source/.vs/
+Source/.vscode/
 .vscode/
 .idea/*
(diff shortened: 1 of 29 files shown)


.gitignore
```diff
@@ -14,5 +14,5 @@ Source/*.tlog
 Source/*.VC.opendb
 Source/*.VC.db
 Source/.vs/
-Source/.vscode/
+.vscode/
```
dreamsyntax, line 17: Please keep both as some open in context of Source folder

dreamsyntax (changes requested):
LGTM; Need to do my own testing with these builds.

dreamsyntax:
Did testing on linux, wow. It runs just as well and less memory (30mb avg vs 100mb avg)
Searching and clearing results uses about the same as expected.
Scrolling and editing in the memory view works just fine.

I'll just need to do some tests on Windows, then once you restore gitignore line I think we're good to merge.

dreamsyntax (approved):
Finished my test on Windows.
15mb vs 75mb (old)
Performance is just as good. - And actually now navigating the memory viewer is much smoother for me.

PR #103, opened 2024-03-05
Reinstate **View > Scanner** top-bar menu action.

### LINKED ISSUE #102: Restore View > Scanner Menu Button
Some users prefer the explicit toggle for enabling/disabling the Scanner rather than dragging the vertical bar.

Restore this button and behavior that was removed in 0.7.0.

cristian64:
Reinstate **View > Scanner** top-bar menu action.

This checkable menu action was removed in dd7a5adab2ba9405d, as it was deemed redundant after the main vertical splitter was added in 76e3b2eb6e77d9e42. However, it turned out that this convenience action was quite popular, and users have requested it back.

An effort has been made to preserve the splitter's position when the scanner's visibility is toggled.

Bonus: The action has been assigned the `S` keyboard shortcut.

Fixes #102.


Source/GUI/MainWindow.cpp
```diff
@@ -74,6 +73,10 @@ void MainWindow::makeMenus()
 
   m_actMemoryViewer = new QAction(tr("&Memory Viewer"), this);
   m_actCopyMemory = new QAction(tr("&Copy Memory Range"), this);
+  m_actScanner = new QAction(tr("&Scanner"), this);
+  m_actScanner->setShortcut(QKeySequence("S"));
```
dreamsyntax, line 77: > Bonus: The action has been assigned the `S` keyboard shortcut.

I'm actually against 'S' being a shortcut for this. It's easy to accidentally trigger, and also prevents the ability of pressing 'S' in the watchlist to snap to the next entry starting with 'S'.

I would be fine with Shift+S or some other combination rather than single key though.

dreamsyntax (changes requested):

Source/GUI/MainWindow.cpp
cristian64: > ...
>
> I would be fine with Shift+S or some other combination rather than single key though.

I'm not particularly happy with `Shift+S`. That kind of shortcut is often used for saving, or saving a copy.

What if we go with `F3` (or some other number?)

dreamsyntax: F3 is fine with me

dreamsyntax: Whatever you decide in the Label under View lets add in parens the key.

Such that that item would say:
`Scanner (F3)`

cristian64: > I think you were thinking Ctrl+Shift+S or Ctrl+S.

Yes, but I wouldn't be surprised if there are other software that use `Shift+S` for similar actions.

Source/GUI/MainWindow.cpp
```diff
@@ -74,6 +73,10 @@ void MainWindow::makeMenus()
 
   m_actMemoryViewer = new QAction(tr("&Memory Viewer"), this);
   m_actCopyMemory = new QAction(tr("&Copy Memory Range"), this);
+  m_actScanner = new QAction(tr("&Scanner"), this);
```
dreamsyntax, line 76: Whatever you decide in the Label under View lets add in parens the key.

Such that that item would say:
`Scanner (F3)`

dreamsyntax (changes requested):

Source/GUI/MainWindow.cpp
cristian64: > Whatever you decide in the Label under View lets add in parens the key.
> 
> Such that that item would say: `Scanner (F3)`

Note that showing the keyboard shortcut in the action's label is something that Qt does automatically by virtue of adding a keyboard shortcut to the `QAction`.

In my system/desktop environment, it shows:
image

To do add the shortcut in parens, we'd have to do modify the action's label ourselves, and then try to find a way so that Qt does not show the keyboard shortcut at the right end.

dreamsyntax: Ah ok, sounds good then. Lets not change it.

dreamsyntax (approved):

PR #105, opened 2024-03-06
Add the `--dolphin-process-name` command-line option.

### LINKED ISSUE #42: Feature: Support any executable name
Dolphin Memory Engine will not hook to Dolphin if the Dolphin executable is renamed. I maintain several Dolphin versions in the same folder and rename them to the version they are and if they are not explicitly named "dolphin.exe", then this tool won't hook to it.

cristian64:
Used to override the name of the Dolphin Emulator process that the DME will search for in the system while trying to hook.

The short version of the option is `-d`.

The [undocumented] `DME_DOLPHIN_PROCESS_NAME` environment variable can
be used alternatively.

Fixes #42.

This work also enables the `--version` and `--help` command-line arguments.

Bonus: Trailing spaces, as well as `\r` characters, have been removed.


cristian64:
The first two commits are only whitespace changes. (I can move them to a separate PR if it makes things easier.)

The changes for Windows and macOS have been made blindly; unable to test. Let's see if the build succeeds.

dreamsyntax:
> Would you able to test at least on Windows? It'd be interesting to test also with non-ASCII characters, which I'm hoping it will work.

Tested with process name '攫'
image

Launching with no arg:
image

Launching with `--dolphin-process-name 攫` or `-d 攫`:
No luck, does not detect the process.

Launching with `--dolphin-process-name 攫.exe` or `-d 攫.exe`:
Attached properly.


Also tested with `ShadowSX` as name of Dolphin process. The results are the same.

--
Testing with no modification to `Dolphin.exe` also works as expected.

--
Can we implicitly add .exe to the search type of the user does not specify? Or do you think its standard to require it?

Source/main.cpp
```diff
@@ -1,10 +1,38 @@
 #include <QApplication>
+#include <QCommandLineParser>
 
 #include "GUI/MainWindow.h"
 
 int main(int argc, char** argv)
 {
   QApplication app(argc, argv);
+  QApplication::setApplicationName("Dolphin Memory Engine");
+  QApplication::setApplicationVersion("0.9.0");
```
dreamsyntax, line 10: Lets set this to 1.0.0 as that is the next release.

Please also add the help/args to the README file.

dreamsyntax (changes requested):

cristian64:
> > Would you able to test at least on Windows? It'd be interesting to test also with non-ASCII characters, which I'm hoping it will work.
> 
> [...]
> 
> -- Can we implicitly add .exe to the search type of the user does not specify? Or do you think its standard to require it?

I think my preference is to be specific, but I'm having a hard time finding a case where the implicit extension could be a problem. So, happy to add it if you haven't changed your mind since your comment.

dreamsyntax:
> > > Would you able to test at least on Windows? It'd be interesting to test also with non-ASCII characters, which I'm hoping it will work.
> > 
> > 
> > [...]
> > -- Can we implicitly add .exe to the search type of the user does not specify? Or do you think its standard to require it?
> 
> I think my preference is to be specific, but I'm having a hard time finding a case where the implicit extension could be a problem. So, happy to add it if you haven't changed your mind since your comment.

I would think
-> check explicit, if not found and on windows assume .exe or something like that

cristian64:
> > > > Would you able to test at least on Windows? It'd be interesting to test also with non-ASCII characters, which I'm hoping it will work.
> > > 
> > > 
> > > [...]
> > > -- Can we implicitly add .exe to the search type of the user does not specify? Or do you think its standard to require it?
> > 
> > 
> > I think my preference is to be specific, but I'm having a hard time finding a case where the implicit extension could be a problem. So, happy to add it if you haven't changed your mind since your comment.
> 
> I would think -> check explicit, if not found and on windows assume .exe or something like that
> 
> Or just make it clear in the help/readme that windows must include the extension. I'm fine with that too.

The option's help description does say the file extension must be given, but the Windows ecosystem is known for not minding.

dreamsyntax (approved):

dreamsyntax:
> > > > > Would you able to test at least on Windows? It'd be interesting to test also with non-ASCII characters, which I'm hoping it will work.
> > > > 
> > > > 
> > > > [...]
> > > > -- Can we implicitly add .exe to the search type of the user does not specify? Or do you think its standard to require it?
> > > 
> > > 
> > > I think my preference is to be specific, but I'm having a hard time finding a case where the implicit extension could be a problem. So, happy to add it if you haven't changed your mind since your comment.
> > 
> > 
> > I would think -> check explicit, if not found and on windows assume .exe or something like that
> > Or just make it clear in the help/readme that windows must include the extension. I'm fine with that too.
> 
> The option's help description does say the file extension must be given, but the Windows ecosystem is known for not minding.

Your call. I'm fine either way, merge when you think its good to go.
If you dont do this:
https://github.com/aldelaro5/dolphin-memory-engine/pull/105#discussion_r1515365623
I can just add it before release.

cristian64:
I've made the tweak so that, on Windows, the `.exe` file extension can be omitted.

dreamsyntax (approved):
Tested with
`-d test`
and
`-d test.exe`

Both work as expected.

PR #106, opened 2024-03-06
Support for multiple value edits.

### LINKED ISSUE #48: Feature: Edit multiple values at once
I'm trying to edit multiple addresses but it just doesn't work.
When i press the value edit shortcut key (F2) when only one address is selected, the value edit is enabled. But if i do the same when multiple addresses are selected, i'm enable to edit them.

cristian64:
When multiple watch entries are selected, and an edit is applied to one of the values, the new value is set in all addresses (provided the watch entry has the same type).

To edit a value without losing the multi selection, `F2` can be pressed while the focus is set on a value cell.

Fixes #48.

### DIFF
--- a/Source/GUI/MemWatcher/MemWatchModel.cpp
+++ b/Source/GUI/MemWatcher/MemWatchModel.cpp
@@ -227,6 +227,12 @@ QVariant MemWatchModel::data(const QModelIndex& index, int role) const
 
 bool MemWatchModel::setData(const QModelIndex& index, const QVariant& value, int role)
 {
+  return editData(index, value, role, true);
+}
+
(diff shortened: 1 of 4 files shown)


dreamsyntax:
In a future PR if not this one,
I think making it a Context Right Click action too should be done so its exposed besides F2

dreamsyntax (approved):
Tested, works good. I think we are good to merge unless you wanted to add the right click context menu addition

cristian64:
> Tested, works good. I think we are good to merge unless you wanted to add the right click context menu addition

I'll look into it later. It can be a different PR. (It's certainly going to be a separate commit anyway.)

PR #108, opened 2024-03-07
Add **Edit Value** action in context menu in main tree view.

cristian64:
Equivalent to pressing `F2` while focusing a cell in the view, or double-clicking the value cell.


dreamsyntax (approved):
Tested. Works great, thanks

PR #111, opened 2024-03-08
Rework of the **About** dialog.

cristian64:
Application logo, as well as links to the source code, license file, contributors page, and releases page, have been included:

Dolphin Memory Engine - About Dialog

### DIFF
--- a/Source/GUI/MainWindow.cpp
+++ b/Source/GUI/MainWindow.cpp
@@ -1,10 +1,10 @@
 #include "MainWindow.h"
 
 #include <QApplication>
+#include <QDialog>
 #include <QHBoxLayout>
 #include <QIcon>
 #include <QMenuBar>
-#include <QMessageBox>
 #include <QShortcut>
 #include <QString>
 #include <QVBoxLayout>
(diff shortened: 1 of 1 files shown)


dreamsyntax (approved):
I'd prefer linking to the project page instead of /releases, but im fine either way.

aldelaro5:
> I'd prefer linking to the project page instead of /releases, but im fine either way.

I am fine either way too, but as a suggestion: you could have a link label for "releases" and another for "source code".

I say this because it is very surprising how users don't know how to access the release page from the home page (even with a link at the start of the readme).

cristian64:
> I'd prefer linking to the project page instead of /releases, but im fine either way.

How about both? What would the label be? And in which position?

cristian64:
Some options:

1. Project Page | License | Contributors | Updates
2. Project Site | License | Contributors | Software Updates
3. License | Contributors | Software Updates
4. License | Contributors | Releases
5. Source Code | License | Contributors | Updates

Thoughts? I don't really have a strong opinion. The application is quite niche: I'm sure anybody that uses it is technical enough to find all the info, even if we don't include it here.

dreamsyntax:
> Some options:
> 
>     1. Project Page | License | Contributors | Updates
> 
>     2. Project Site | License | Contributors | Software Updates
> 
>     3. License | Contributors | Software Updates
> 
>     4. License | Contributors | Releases
> 
>     5. Source Code | License | Contributors | Updates
> 
> 
> Thoughts? I don't really have a strong opinion. The application is quite niche: I'm sure anybody that uses it is technical enough to find all the info, even if we don't include it here.

I like
1 or 5

Open to either, your choice.

cristian64:
I've gone for Option 5. (Happy to make further tweaks if more suggestions are made.)

dreamsyntax (approved):

cristian64:
As I was checking how the dialog looked like on Windows, I noticed the icon is not working:

image

Then I noticed the build is not bundling the SVG plug-in. The external Qt library does not have it either; it appears to be a reduced version. We may have to get a new, more complete version (perhaps Qt 6.5.3).

cristian64:
SVG logo is now rasterized after #112:

image

Now noticing that the copyleft symbol 🄯 (ɔ) is not available on the font family on Windows. (I may open a PR for this.)

---

We also don't know how the dialog looks like on macOS. As mentioned in my previous comment, the external Qt library for macOS in the Git submodule does not include the QtSvg component; I guess the SVG logo won't work there until that is addressed.

@campital Is this something you would like to look into? When you uploaded the macOS library, did you leave out the QtSvg module, or was it not compiled by default in macOS?

PR #112, opened 2024-03-09
Configure CMake to install Qt6Svg.dll and qsvg.dll in build.

cristian64:
Without these libraries, SVG icon engine is not available, and SVG images (only application logo at this time) cannot be rasterized.


dreamsyntax (approved):
Untested, but looks correct

PR #117, opened 2024-04-25
MemWatchEntry: Show address pointed to by the chain

malleoz:
I am currently fleshing out a DME watch table for various classes, each having several members. These classes are allocated dynamically, so their location in memory changes. Often times, I will want to use DME to quickly resolve the pointer chains so I know where to set memory breakpoints in Dolphin for particular members. The current address column behavior does not give the user any information unique to that entry; two entries using the same pointer but with different offsets show the same address information. In order to deduce where in memory these members are, I have to double-click the entry to see the entire pointer chain in the pop-up window.

**Before:**

**After:**

### DIFF
--- a/Source/GUI/MemWatcher/MemWatchModel.cpp
+++ b/Source/GUI/MemWatcher/MemWatchModel.cpp
@@ -200,9 +200,7 @@ QVariant MemWatchModel::data(const QModelIndex& index, int role) const
       }
       case WATCH_COL_ADDRESS:
       {
-        u32 address = entry->getConsoleAddress();
-        bool isPointer = entry->isBoundToPointer();
-        return getAddressString(address, isPointer);
+        return getAddressString(entry);
       }
       case WATCH_COL_VALUE:
       {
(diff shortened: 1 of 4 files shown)


Source/GUI/MemWatcher/MemWatchModel.cpp
```diff
@@ -624,14 +622,17 @@ void MemWatchModel::sortRecursive(int column, Qt::SortOrder order, MemWatchTreeN
...
 
-QString MemWatchModel::getAddressString(u32 address, bool isPointer) const
+QString MemWatchModel::getAddressString(const MemWatchEntry* entry) const
 {
   std::stringstream ss;
-  if (isPointer)
+  if (entry->isBoundToPointer())
   {
-    ss << "P->" << std::hex << std::uppercase << address;
+    int level = static_cast<int>(entry->getPointerLevel());
```
dreamsyntax, line 630: What was the resolution on the discussion on static_cast here?

malleoz: This is separate to the discussion on Discord, where @cristian64 pointed out int-to-float conversions cannot use `static_cast` due to aliasing. However, in this scenario we are casting from `size_t` to `int` which is safe and is used various other times in `DlgAddWatchEntry.cpp` when retrieving the pointer level.

dreamsyntax (approved):
Tested on windows. LGTM.

PR #122, opened 2024-04-28
Show extended scan preview for strings.

### LINKED ISSUE #79: Feature: Extended Scan Result Preview for Strings
* Add a way to preview longer strings than searched.

## Example

I search "cw" and retrieve a list of strings.

image

I want to see the remainder of the string until the next null terminator, or a settable number of characters to attempt to print.

If 80233ED1 was "cwlater!", I would want to see that in the scan results.

cristian64:
Scan preview for strings will be unbounded; their length will be determined solely by the null character:

DME - Extended Preview Strings
<p></p>

Bonus: Non-printable characters in scanned or watched strings will be replaced with the � character (the unrepresentable character).

Fixes #79.

### DIFF
--- a/Source/Common/MemoryCommon.cpp
+++ b/Source/Common/MemoryCommon.cpp
@@ -1,6 +1,7 @@
 #include "MemoryCommon.h"
 
 #include <bitset>
+#include <cctype>
 #include <cstring>
 #include <iomanip>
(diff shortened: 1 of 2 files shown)


dreamsyntax (approved):
Awesome, this is super useful

PR #125, opened 2024-05-04
Add Clang-Tidy configuration file + a some fixes.

cristian64:
A `.clang-tidy` file has been added to the repository (next to the `.clang-format` file).

A subset of the enabled warning types have been addressed.


cristian64:
Discovered the hard way that [`std::bit_cast`](https://en.cppreference.com/w/cpp/numeric/bit_cast) is not available in the C++20 feature subset that GCC 10 implements. I am on GCC 11 in my system, so I didn't notice. A hand-rolled implementation has been added; checks are now passing.

dreamsyntax:
Do you have a preference on if this is squashed into one commit on merge, or prefer them rebased onto the repo as is?

cristian64:
My preference is to rebase and merge without squashing (it'd be a pity to lose the info I put in the description in each of the commits).

dreamsyntax (approved):

PR #126, opened 2024-05-05
Support array-of-bytes binary display

malleoz:
Viewing an `array of bytes` entry as binary can be especially useful when observing bitfields. For example, Mario Kart Wii uses bitfields to track whether we are accelerating, braking, drifting, tricking, etc. Being able to view each bit in the field makes viewing the current state of each scenario a bit easier!

I didn't add octal or decimal support, because I can't imagine any scenarios where such a display could be useful.

### DIFF
--- a/Source/Common/MemoryCommon.cpp
+++ b/Source/Common/MemoryCommon.cpp
@@ -405,6 +405,11 @@ char* formatStringToMemory(MemOperationReturnCode& returnCode, size_t& actualLen
       }
       else
       {
+        if (base == MemBase::base_binary && next.size() == 8)
+        {
+          bytes.push_back(next);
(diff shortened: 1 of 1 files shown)


cristian64:
I've given it a go, and it works well. I've added a few comments, but I wouldn't be against to get it merged in its current state.

(The macOS error appears to be unrelated; might need to retry.)

malleoz:
@cristian64 I don't see your comments. Do you have a pending review that you forgot to click "Submit review" for?

malleoz:
Those changes seem to work fine on my end, so I'm all for cleaning it up!

dreamsyntax (approved):

PR #127, opened 2024-05-05
Update VS project to C++20

malleoz:
ae32ba2 introduces a C++20 standard, which exposes that the Visual Studio project was not updated to support the newer standard.

### DIFF
--- a/Source/dolphin-memory-engine.vcxproj
+++ b/Source/dolphin-memory-engine.vcxproj
@@ -57,7 +57,7 @@
       <AdditionalIncludeDirectories>$(QTDIR)include;$(QTDIR)include\QtCore;$(QTDIR)include\QtGui;$(QTDIR)include\QtWidgets</AdditionalIncludeDirectories>
       <ObjectFileName>$(IntDir)%(RelativeDir)</ObjectFileName>
       <PreprocessorDefinitions>WIN32;WIN64;QT_DLL;QT_CORE_LIB;QT_GUI_LIB;QT_WIDGETS_LIB;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <LanguageStandard Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">stdcpp17</LanguageStandard>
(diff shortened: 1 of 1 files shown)


cristian64:
Looks good to me.

CMake file had been updated to C++20 in a0bb1a4b86d19ddc10b6b543bf3a3b15203f092e (#120), but the VS project files were overlooked. (Sorry!)

Source/dolphin-memory-engine.vcxproj
```diff
@@ -57,7 +57,7 @@
       <AdditionalIncludeDirectories>$(QTDIR)include;$(QTDIR)include\QtCore;$(QTDIR)include\QtGui;$(QTDIR)include\QtWidgets</AdditionalIncludeDirectories>
       <ObjectFileName>$(IntDir)%(RelativeDir)</ObjectFileName>
       <PreprocessorDefinitions>WIN32;WIN64;QT_DLL;QT_CORE_LIB;QT_GUI_LIB;QT_WIDGETS_LIB;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <LanguageStandard Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">stdcpp17</LanguageStandard>
+      <LanguageStandard Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">stdcpp20</LanguageStandard>
       <LanguageStandard_C Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">stdc17</LanguageStandard_C>
```
dreamsyntax, line 61: Do we care to have this declared / should we update it for the heck of it?

dreamsyntax: Actually, it looks like `LanguageStandard_C` is ignored if LanguageStandard is defined anyway. So we can remove that line entirely.

cristian64: Note that `stdc20` is not a thing at this time, I think. Whether the field can be removed, I don't know.

dreamsyntax: Considering its not related to what C++ ver we are using, no reason to address it in this PR.

dreamsyntax (approved):

Issue #119, opened 2024-04-26
Maintenance: Resolve Build Warnings (Windows)


dreamsyntax:
From Windows build output:

```
D:\a\dolphin-memory-engine\dolphin-memory-engine\Source\DolphinProcess\Windows\WindowsDolphinProcess.cpp(20,29): warning C4996: 'std::codecvt_utf8<wchar_t,1114111,(std::codecvt_mode)0>': warning STL4017: std::wbuffer_convert, std::wstring_convert, and the <codecvt> header (containing std::codecvt_mode, std::codecvt_utf8, std::codecvt_utf16, and std::codecvt_utf8_utf16) are deprecated in C++17. (The std::codecvt class template is NOT deprecated.) The C++ Standard doesn't provide equivalent non-deprecated functionality; consider using MultiByteToWideChar() and WideCharToMultiByte() from <Windows.h> instead. You can define _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to suppress this warning. [D:\a\dolphin-memory-engine\dolphin-memory-engine\Source\bin\dolphin-memory-engine.vcxproj]
  C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.38.33130\include\codecvt(34,1):
  see declaration of 'std::codecvt_utf8'
  
[...]

PR #128, opened 2024-05-06
Enable compile warnings.

cristian64:
- `-Wall` and `-Wextra` have been enabled in GCC/Clang.
- `-W4` has been enabled in MSVC.

All warnings have been promoted to errors on Unix compilers via `-Werror`. All warnings (with a couple of exceptions in Clang) have been addressed.

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -61,6 +61,31 @@ IF(WIN32)
 SET(GUI_TYPE WIN32)
 ENDIF(WIN32)
 
+if (UNIX)
+    add_compile_options(
+        -Wall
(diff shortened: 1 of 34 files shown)


dreamsyntax:
https://github.com/aldelaro5/dolphin-memory-engine/issues/119

dreamsyntax (approved):

cristian64:
> #119

Sorry; forgot to comment on this. This PR was only to address warnings in Unix systems, so #119 has not been addressed as part of this. (In fact, now there are even more warnings that need to be addressed on MSVC, as `/W4` was been enabled.)

PR #134, opened 2024-05-12
Set initial state of the **Scanner** checkable action in the next event loop tick.

### LINKED ISSUE #131: Bug: Initial "Scanner" visible check/unchecked state in wrong state
The "Scanner" on a fresh install / no config of DME will incorrectly have the Scanner checkbox unchecked while the Scanner is shown, thus clicking it or pressing F3 will need to be done twice to hide it properly.

image

cristian64:
Previously, the initial size of the main splitter was used to initialize the checkable action. However, that size is not accurate until the main window has been shown.

The action is now initialized in the next event loop tick, when the GUI is ready.

Fixes #131.

### DIFF
--- a/Source/GUI/MainWindow.cpp
+++ b/Source/GUI/MainWindow.cpp
@@ -7,6 +7,7 @@
 #include <QMenuBar>
 #include <QShortcut>
 #include <QString>
+#include <QTimer>
 #include <QVBoxLayout>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


dreamsyntax (approved):

Source/GUI/MainWindow.cpp
```diff
@@ -80,7 +81,10 @@ void MainWindow::makeMenus()
   m_actScanner = new QAction(tr("&Scanner"), this);
   m_actScanner->setShortcut(QKeySequence("F3"));
   m_actScanner->setCheckable(true);
-  m_actScanner->setChecked(m_splitter->sizes()[0] > 0);
+  QTimer::singleShot(0, [this]() {
```
dreamsyntax, line 84: I'm assuming we dont have a way around not using QTimer?

cristian64: 1. Another option would be to connect to the `aboutToShow` signal of the menu, and update its state every time.
2. Another option could be to force the visibility of the main window within the `MainWindow()` constructor [provided this can be done within the constructor], and then update the action's state before leaving the constructor.
3. Another option could be to check whether it's a fresh config and, if so, set the state to `true`, as we'd know no setting that would ultimately reset the splitter could be called.

I'm not sure 2. and 3. would be very elegant. 1. could work, provided the state of the action [which would be inconsistent until the menu is shown] is not relied upon [now or in the future].

(I'd probably stick with the single-shot timer to be honest.)

dreamsyntax: I'm not well versed in Qt so I'll trust your judgement with the singleShot then

PR #135, opened 2024-05-12
Warn user about other instances owning the settings file.

cristian64:
When multiple DME instances are created, only the first one gets to own the settings file now; previously, the last DME instance to close would rewrite the settings file, potentially causing data loss from the previous instances.

A warning dialog will be shown on startup if the settings lockfile cannot be acquired. On exit, if there are unsaved changes to the watch list, the user will be given an opportunity to save it into a separate `.dmw` file.


dreamsyntax (approved):

dreamsyntax (approved):
Tested on Windows post rebase.

PR #138, opened 2024-05-14
Address `-Wfortify-source` warnings.

cristian64:
A number of `constexpr` checks have been added to the `compareMemoryAsNumbersWithType<T>()` template function so that the compiler can determine that type conversions are only performed between compatible types (i.e. types of the same width).

In addition, `Common::bit_cast` is now used for type conversion (instead of raw `memcpy()`).

`-Wfortify-source`; previously suppressed; has been re-enabled now. Note that this is a Clang-specific warning type.

### DIFF
--- a/Source/MemoryScanner/MemoryScanner.h
+++ b/Source/MemoryScanner/MemoryScanner.h
@@ -75,54 +75,45 @@ public:
     T secondByte;
     std::memcpy(&firstByte, first, sizeof(T));
     std::memcpy(&secondByte, second, sizeof(T));
-    size_t size = sizeof(T);
-    switch (size)
+
+    constexpr size_t size{sizeof(T)};
+    static_assert(size == 1 || size == 2 || size == 4 || size == 8, "Unexpected type size");
+
+    if constexpr (size == 2)
     {
-    case 2:
-    {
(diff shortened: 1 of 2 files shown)


Source/MemoryScanner/MemoryScanner.h
```diff
@@ -75,54 +75,45 @@ class MemScanner
...
-        std::memcpy(&firstDoubleword, &secondByte, sizeof(u64));
+        firstDoubleword = Common::bit_cast<u64, T>(secondByte);
         firstDoubleword = Common::bSwap64(firstDoubleword);
-        std::memcpy(&secondByte, &firstDoubleword, sizeof(u64));
+        secondByte = Common::bit_cast<T, u64>(firstDoubleword);
       }
-      break;
-    }
-    default:
-      assert(0 && "Unexpected type size");
```
cristian64, line 124: This `assert` was introduced in e9c0e22ed318daa3d, and it was _not_ correct. `T` could be `char`, and it's a valid case where it simply does not require any byte swapping in any case (for obvious reasons).

I only discovered this after converting the `assert` to `static_assert` (moved towards the top of the function now).

dreamsyntax (approved):

PR #144, opened 2024-05-18
Remove remnants of the deactivated sorting mechanism.

cristian64:
This sorting mechanism was deactivated in 65bb529fbd6f52ac547e, but the actual functionality was still in the code base.

Note that sorting was disabled due to its implementation being destructive: the actual watch nodes were reordered; not just the visual representation of the items in the Qt tree view. If the functionality is reinstated in the future, a `QSortFilterProxyModel` should be used.


dreamsyntax (approved):

dreamsyntax:
We probably wont ever want this feature, but a search filter that doesnt destroy the dmw struct could be cool.

PR #146, opened 2024-05-18
Treat failed save actions as cancel actions.

cristian64:
Previously, if the user exited the file browser dialog without saving the file (voluntarily, or due to I/O error), DME would still exit (or wipe out the watch model; depending on the case), causing unrecoverable data loss.

Test plan:
- Launch DME.
- Add a few watch entries to mark the model with unsaved changes.
- Clear the watch list via the **File > Clear the watch list** action.
- A confirmation dialog with the **Cancel**, **No**, and **Yes** options should appear.
- Click **Yes**
- Once the file browser dialog opens, close it without selecting a file (or choose a filepath where the user does not have write permissions).

Previous behavior: The watch list was wrongly cleared, without the user having saved it.

New behavior: The user gets back to DME, as if **Cancel** had been selected in the confirmation dialog.

### DIFF
--- a/Source/GUI/MemWatcher/MemWatchWidget.cpp
+++ b/Source/GUI/MemWatcher/MemWatchWidget.cpp
@@ -672,7 +672,7 @@ void MemWatchWidget::openWatchFile()
   }
 }
 
-void MemWatchWidget::saveWatchFile()
+bool MemWatchWidget::saveWatchFile()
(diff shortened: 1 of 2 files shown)


dreamsyntax (approved):
Good catch. I'm sure I've been burned by this before without realizing it.

PR #147, opened 2024-05-18
Prompt user before destructive operations of great impact.

cristian64:
The **Clear the watch list** action did not always show a confirmation dialog, only when there were unsaved changes, which overlooked the case where the watch list is only backed up by the data stored in the settings file.

### DIFF
--- a/Source/GUI/MemWatcher/MemWatchModel.cpp
+++ b/Source/GUI/MemWatcher/MemWatchModel.cpp
@@ -186,7 +186,7 @@ void MemWatchModel::clearRoot()
   endResetModel();
 }
 
-void MemWatchModel::removeNode(const QModelIndex& index)
+void MemWatchModel::deleteNode(const QModelIndex& index)
 {
   if (index.isValid())
   {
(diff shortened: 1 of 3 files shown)


dreamsyntax:
I dislike having a prompt for Cut.

If we want this it should be a configurable Option state that can be enabled/disabled

"Confirm Cut" or "Confirm on Cut" etc as the name

aldelaro5:
Yeah I am not a big fan of confirm on cut either.

For clear list, it's reasonable (maybe it can offer a save before, but I am fine either way). It's just cut is one that you may do very frequently enough.

cristian64:
> Yeah I am not a big fan of confirm on cut either.
> 
> For clear list, it's reasonable (maybe it can offer a save before, but I am fine either way). It's just cut is one that you may do very frequently enough.

I guess the data is still in the clipboard, so no need to bother the user when cutting.

dreamsyntax:
> > Yeah I am not a big fan of confirm on cut either.
> > 
> > For clear list, it's reasonable (maybe it can offer a save before, but I am fine either way). It's just cut is one that you may do very frequently enough.
> 
> I guess the data is still in the clipboard, so no need to bother the user when cutting.

Just an additional data point most modern OS/DE ship clipboard history by default now too, so id be fine not having that feature at all.

cristian64:
I've now dropped the changes for the cut action.

dreamsyntax (approved):

PR #148, opened 2024-05-18
Add icon to watch groups.

cristian64:
A folder icon is used. For empty watch groups, the folder icon is empty.


dreamsyntax (commented):
~~Looks good, though we should add a credits line in About at the very least~~

Update: cristian64 created a new icon in Inkscape instead of using the other licensed icons, so redundant.

dreamsyntax (approved):
tested on linux

PR #149, opened 2024-05-18
Add Light, Dark Gray, and Dark themes along with the System theme.

cristian64:
The following themes are offered:

- **System**: Purely based on Qt's Fusion built-in style
- **Light**: Qt's Fusion built-in style + custom light palette
- **Dark Gray**: Qt's Fusion built-in style + custom dark gray palette
- **Dark**: Qt's Fusion built-in style + custom dark palette

| Light | Dark Gray | Dark |
| --- | --- | --- |
| DME - Light Theme | DME - Dark Gray Theme | DME - Dark Theme |

### DIFF
--- a/Source/GUI/GUICommon.cpp
+++ b/Source/GUI/GUICommon.cpp
@@ -2,7 +2,9 @@
 
 #include <QApplication>
 #include <QCoreApplication>
+#include <QFontMetrics>
 #include <QStringList>
+#include <QToolTip>
 
 namespace GUICommon
 {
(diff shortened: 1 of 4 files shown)


dreamsyntax:
Lint failed.

Im fine with this if we extend to be part of the theme switcher which is currently Windows only.

Since default linux/mac already has dark mode iff using appropriate Qt theme selection on your system.

dreamsyntax:
Light:
image

Dark:
image

--

I think for Linux/Mac platforms we should probably specify that 'Light' is actually your default Qt theme- so in my case of using EndeavourOS Cinnamon with Dark Mode Qt / Dark themes looks as above for each option.

cristian64:
> [...]
> 
> I think for Linux/Mac platforms we should probably specify that 'Light' is actually your default Qt theme- so in my case of using EndeavourOS Cinnamon with Dark Mode Qt / Dark themes looks as above for each option.

This is the part I mentioned I'd like to discuss. Right now, the naked use of the Fusion built-in style means "use the system's". Isn't that the case on Windows too? I would have thought a user having the light theme on Windows would still see a light theme in DME when the current Dark mode is selected.

We could offer the following options for the **Theme** setting:

- Default (or System): the Fusion built-in style as-is
- Dark: Fusion + custom dark palette
- Light: Fusion + custom light palette

This would guarantee that the Dark and Light styles look alike in all platforms.

dreamsyntax:
> Right now, the naked use of the Fusion built-in style means "use the system's". Isn't that the case on Windows too? I would have thought a user having the light theme on Windows would still see a light theme in DME when the current Dark mode is selected.

You are correct, just tested this. It does follow system theme.

dreamsyntax:
Minor nitpick, can we make the names expose the theme names? Like how DuckStation/PCSX2 do it?

The system dark on Windows for example is Fusion Dark,
Where the "Dark" option is actually Fusion Gray


image

image

DuckStation/PCSX2 example:
image

dreamsyntax (approved):
LGTM.
Note this will drop WindowsVistaTheme, but I think it doesn't look that different compared to Fusion Light anyway.

PR #150, opened 2024-05-18
Attempt to make the directories before trying to lock the lockfile.

cristian64:
In a fresh installation where the `.ini` file does not yet exist, the application could wrongly determine that another instance was running as the lockfile could not be created due to the missing parent directory.

All the intermediate-level directories needed to contain the lockfile are now created if they did not exist.


dreamsyntax (approved):
Confirmed working on linux

PR #153, opened 2024-05-26
Add memory address input custom widget.

cristian64:
The widget features a built-in "0x" prefix, and a validator that validates the user input, removing any additional "0x" prefix. Length is limited to 8 characters, and it uses the system's fixed-width font.

DME - Address Input Widget

The custom widget is used in the following components:
- **Add/Edit Watch** dialog
- **Copy Memory Range** dialog
- **Memory Viewer** dialog
- **Search Range** fields


dreamsyntax (approved):
Tested and verified strings of n size are not interrupted.

PR #155, opened 2024-05-26
CI: move lint into one workflow

dreamsyntax:
* Moves the lint job into build.yml so there is only one clickable workflow per PR. I was getting annoyed accidentally clicking lint and then having to go back and renavigate to build when pulling artifacts.


cristian64:
I don't quite remember what motivated having it separated into two.  Was my intention to add further linter checks (Clang-Tidy or Cppcheck), or was it just a copy of something I had seen in other projects [and that I won't remember now].

Either way, it doesn't make a big deal.

PR #156, opened 2024-05-27
Move Dolphin hook status to status bar.

cristian64:
It reduces the clutter in the GUI by saving several lines of vertical space in the main window.

Bonus: A state icon is now included for each of the four possible states, as well as the game ID of the emulated game when present.

DME - Unhooked
DME - Dolphin not detected
DME - No game running
DME - Hooked


dreamsyntax:
Hmm... I know its traditional to have the status bar on the bottom, but would you consider having an option for the top? I may add it as an option under View.

Functionality tested and the spacing is nice, but there's just something with muscle memory with me wanting the bar up top like this mockup:

image

dreamsyntax (approved):
Approving as I think its just a me thing, and I can add the above later.

cristian64:
In the mockup, the status bar is in a place that is occupied by a label [that is only visible after the first search]: the one that shows number of matches.

I too noticed muscle memory was a thing, but to be fair it went away after just a few minutes. 

One option we can consider (I'd love to see mockups) is moving it to the very right in the top menu's bar (at the very right of the **Help** menu). We might need to reverse the layout direction so that the icon stays at the corner.

PR #158, opened 2024-05-27
Persistent expansion state.

cristian64:
The `.dmw` file format now includes the expansion state of group nodes. When the file is loaded, the expansion state is restored. This applies to copy-pasted rows.


dreamsyntax (approved):
Tested with change for nulls, worked good on Windows with expected diff results

PR #159, opened 2024-05-29
Use correct font in painter in **Memory Viewer** dialog.

cristian64:
Although the font was set up in the **Memory Viewer** dialog's main widget, the viewport's painter that actually renders the text was still using the default font.

When a theme (other than **System**) was used, this resulted in the grid and the text having a mismatched size, which was specially noticeable when the font size was made large or small:

| Before | After |
| --- | --- |
| DME - Memory Viewer (incorrect font size) png | DME - Memory Viewer (correct font size) |

Bonus changes: initial font size is no longer hardcoded to 15; now it's based on a percentage (+50%) of the size of the default font.

### DIFF
--- a/Source/GUI/MemViewer/MemViewer.cpp
+++ b/Source/GUI/MemViewer/MemViewer.cpp
@@ -372,15 +372,9 @@ void MemViewer::updateFontSize(int newSize)
 {
   m_memoryFontSize = newSize;
 
-#ifdef __linux__
-  setFont(QFont("Monospace", m_memoryFontSize));
-#elif _WIN32
-  setFont(QFont("Courier New", m_memoryFontSize));
-#elif __APPLE__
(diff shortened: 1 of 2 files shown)


dreamsyntax:
I'm not sure what was going on with your before screenshots... I see no change on Windows at least.

Windows Before:
image

Windows After:
image

---
I will test on linux shortly

dreamsyntax (approved):
The changes seem sane to me, I just don't notice any difference in my tests on both windows and linux (cachyos, kde, wayland)

If it fixes this issue for some other configurations, I don't have a problem with it being merged.
Feel free to merge when you are satisfied.

dreamsyntax:
> I have now realized that the issue only manifests if the theme is not set to **System**. Could you confirm you can reproduce the problem with any other theme?

Ah yep, thats what it was.
Reproduced on Windows
image

PR #161, opened 2024-06-07
Relax input acceptance criteria in `AddressInputWidget`.

cristian64:
The max length (previously `10`) has been lifted, and the input is now self-corrected as the user enters text:

- Invalid characters are removed from the input.
- The optional `0x` prefix is removed.
- Any excess beyond the 8th character is removed.

For example, this now allows the user to paste an address that may contain spaces, which is a fairly common case (e.g. `80 12 34 56`).


dreamsyntax (approved):
seems reasonable, untested

Issue #160, opened 2024-05-31
Feature: Copy address of pointer level


dreamsyntax:
For pointers of n depth, you can "Browse memory at" to open the Memory Viewer to this address.

![image](https://github.com/aldelaro5/dolphin-memory-engine/assets/14857235/6c3132ae-10cb-4bd4-88bd-571849c7e957)

Currently there is no way to 'Copy' a particular depth's address. For example, given the below image:
![image](https://github.com/aldelaro5/dolphin-memory-engine/assets/14857235/4ac5973d-d1a4-4110-891f-19cc96127b7a)

If I wanted to copy level 2's address for use in Dolphin's Breakpoint feature I have to manually open this Address edit pane and type it myself in Dolphin, or utilize the "Browse memory at" and then right click -> add to watch -> copy to clipboard rather than hitting ok -> closing the dialog.

This has the potential to be made much simpler by either allowing copying directly from the Address edit pane or adding a dialog option when right clicking an address in the watchlist.

jahorta:
So I took a shot at this, but I would like some feedback before making a PR. I implemented a right click menu entry for copying the address, but I am unsure if I put it in a good position in the right click menu. Currently I have it below the copy option (see below), but it might also work below the browse at address option. Thoughts?

When right clicking a non-pointer
![Sample copy addresses non-pointer](https://github.com/aldelaro5/dolphin-memory-engine/assets/8341369/7c15d804-bbea-417f-9dbb-6e767457ed63)

When right clicking a pointer
![Sample Copy Addresses Pointer](https://github.com/aldelaro5/dolphin-memory-engine/assets/8341369/3e6cd0a1-7bed-4a41-a4b4-e4d1826823bf)

dreamsyntax:
> So I took a shot at this, but I would like some feedback before making a PR. I implemented a right click menu entry for copying the address, but I am unsure if I put it in a good position in the right click menu. Currently I have it below the copy option (see below), but it might also work below the browse at address option. Thoughts?
> 
> When right clicking a non-pointer
> ![Sample copy addresses non-pointer](https://github.com/aldelaro5/dolphin-memory-engine/assets/8341369/7c15d804-bbea-417f-9dbb-6e767457ed63)
> 
> When right clicking a pointer
> ![Sample Copy Addresses Pointer](https://github.com/aldelaro5/dolphin-memory-engine/assets/8341369/3e6cd0a1-7bed-4a41-a4b4-e4d1826823bf)
[...]

PR #167, opened 2024-07-07
Add options to copy pointer addresses (or base address) of a watch entry

jahorta:
Added two methods for copying a pointer address to the clipboard

- Added a context menu to the Edit Address widget for the pointer widget to copy an address from the address label

- Added an option ("Copy address") to the context menu of a Memory Watch widget to copy either the base address of the memory watch, or a selected pointer level if the watch is a pointer.

The second item is currently located after the "Copy" entry of the context menu, though it might fit better after the "Browse memory at" entry. 

reference: #160

### DIFF
--- a/Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
+++ b/Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
@@ -446,7 +446,7 @@ void DlgAddWatchEntry::onPointerOffsetContextMenuRequested(const QPoint& pos)
     if (0 < yPos && yPos < lbl->height() && 0 < xPos && xPos < lbl->width())
     {
       QAction* copyAddr = new QAction(tr("&Copy Address"), this);
-      connect(copyAddr, &QAction::triggered, this, [this, lbl] {
+      connect(copyAddr, &QAction::triggered, this, [lbl] {
         QApplication::clipboard()->setText(lbl->text().mid(4, lbl->text().length() - 4));
       });
       contextMenu->addAction(copyAddr);
--- a/Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
+++ b/Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
@@ -440,7 +440,7 @@ MemWatchEntry* DlgAddWatchEntry::stealEntry()
 
 void DlgAddWatchEntry::onPointerOffsetContextMenuRequested(const QPoint& pos)
 {
-  QLabel* const lbl = static_cast<QLabel*>(sender());
+  QLabel* const lbl = qobject_cast<QLabel*>(sender());
 
   QMenu* contextMenu = new QMenu(this);
   QAction* copyAddr = new QAction(tr("&Copy Address"), this);
--- a/Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
+++ b/Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
@@ -169,7 +169,7 @@ void DlgAddWatchEntry::fillFields(MemWatchEntry* entry)
         m_offsets.append(txbOffset);
         QLabel* lblAddressOfPath = new QLabel();
         lblAddressOfPath->setText(
-          QString::fromStdString(" -> " + m_entry->getAddressStringForPointerLevel(i + 1)));
+            QString::fromStdString(" -> " + m_entry->getAddressStringForPointerLevel(i + 1)));
         lblAddressOfPath->setProperty("addr", m_entry->getAddressForPointerLevel(i + 1));
         lblAddressOfPath->setContextMenuPolicy(Qt::CustomContextMenu);
         connect(lblAddressOfPath, &QWidget::customContextMenuRequested, this,
@@ -198,10 +198,11 @@ void DlgAddWatchEntry::addPointerOffset()
   m_offsets.append(txbOffset);
   QLabel* lblAddressOfPath = new QLabel(" -> ");
   lblAddressOfPath->setText(
-    QString::fromStdString(" -> " + m_entry->getAddressStringForPointerLevel(level + 1)));
+      QString::fromStdString(" -> " + m_entry->getAddressStringForPointerLevel(level + 1)));
   lblAddressOfPath->setProperty("addr", m_entry->getAddressForPointerLevel(level + 1));
   lblAddressOfPath->setContextMenuPolicy(Qt::CustomContextMenu);
-  connect(lblAddressOfPath, &QWidget::customContextMenuRequested, this, &DlgAddWatchEntry::onPointerOffsetContextMenuRequested);
+  connect(lblAddressOfPath, &QWidget::customContextMenuRequested, this,
+          &DlgAddWatchEntry::onPointerOffsetContextMenuRequested);
   m_addressPath.append(lblAddressOfPath);
   m_offsetsLayout->addWidget(lblLevel, level, 0);
   m_offsetsLayout->addWidget(txbOffset, level, 1);
@@ -397,9 +398,9 @@ void DlgAddWatchEntry::updatePreview()
     for (int i = 0; i < level; ++i)
     {
       QLabel* lblAddressOfPath =
-        qobject_cast<QLabel*>(m_offsetsLayout->itemAtPosition(i, 2)->widget());
+          qobject_cast<QLabel*>(m_offsetsLayout->itemAtPosition(i, 2)->widget());
       lblAddressOfPath->setText(
-        QString::fromStdString(" -> " + m_entry->getAddressStringForPointerLevel(i + 1)));
+          QString::fromStdString(" -> " + m_entry->getAddressStringForPointerLevel(i + 1)));
       lblAddressOfPath->setProperty("addr", m_entry->getAddressForPointerLevel(i + 1));
     }
   }
@@ -447,7 +448,7 @@ void DlgAddWatchEntry::onPointerOffsetContextMenuRequested(const QPoint& pos)
 
   const QString text{QString::number(lbl->property("addr").toUInt(), 16).toUpper()};
   connect(copyAddr, &QAction::triggered, this,
-    [text] { QApplication::clipboard()->setText(text); });
+          [text] { QApplication::clipboard()->setText(text); });
   contextMenu->addAction(copyAddr);
 
   if (!lbl->property("addr").toUInt())
--- a/Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
+++ b/Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
@@ -61,9 +61,6 @@ void DlgAddWatchEntry::initialiseWidgets()
   connect(m_btnRemoveOffset, &QPushButton::clicked, this, &DlgAddWatchEntry::removePointerOffset);
 
   m_pointerWidget = new QGroupBox("Offsets (in hex)", this);
-  m_pointerWidget->setContextMenuPolicy(Qt::CustomContextMenu);
-  connect(m_pointerWidget, &QWidget::customContextMenuRequested, this,
-          &DlgAddWatchEntry::onPointerOffsetContextMenuRequested);
 
   m_txbLabel = new QLineEdit(this);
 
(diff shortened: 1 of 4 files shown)


Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
```diff
@@ -424,3 +430,32 @@ MemWatchEntry* DlgAddWatchEntry::stealEntry()
...
+      contextMenu->addAction(copyAddr);
+      if (lbl->text().contains(QString("??")))
+      {
+        copyAddr->setEnabled(false);
+      }
+      contextMenu->popup(lbl->mapToGlobal(click_pos));
+      break;
+    }
+  }
+}
```
cristian64, line 461: Question: Shouldn't we be able to register the context menu to the label directly to avoid the shenanigans with the layout, and all the magic numbers and magic strings?

Suggestion:

When the label widgets are inserted in the layout, a Qt property (see [`QObject::setProperty()`](https://doc.qt.io/qt-6/qobject.html#setProperty)) would be set to hold the required information (perhaps an index, or perhaps the memory offset itself). Then, when the label widget is right-clicked, and the context menu is shown, the slot would check the `sender()`'s property (see [`QObject::property()`](https://doc.qt.io/qt-6/qobject.html#property)).

Source/GUI/MemWatcher/MemWatchWidget.cpp
```diff
@@ -273,6 +273,43 @@ void MemWatchWidget::onMemWatchContextMenuRequested(const QPoint& pos)
   connect(copy, &QAction::triggered, this, [this] { copySelectedWatchesToClipBoard(); });
   contextMenu->addAction(copy);
 
+  if (index != QModelIndex())
```
cristian64, line 276: Question: Why not test whether `index` is valid instead?

```diff
- if (index != QModelIndex())
+ if (index.isValid())
```

Source/GUI/MemWatcher/MemWatchWidget.cpp
```diff
@@ -273,6 +273,43 @@ void MemWatchWidget::onMemWatchContextMenuRequested(const QPoint& pos)
...
+  {
+    MemWatchEntry* entry = m_watchModel->getEntryFromIndex(index);
```
cristian64, line 278: Nit: `entry` should be marked as `const`:
```diff
- MemWatchEntry* entry = m_watchModel->getEntryFromIndex(index);
+ MemWatchEntry* const entry = m_watchModel->getEntryFromIndex(index);
```

Same goes for a few of the other points below such as `copyAddrSubmenu`, `copyPointer`, ... Basically, all variables that are not expected to change should be marked as `const`, including points, parameters, and whatnot.

Source/GUI/MemWatcher/MemWatchWidget.cpp
```diff
@@ -273,6 +273,43 @@ void MemWatchWidget::onMemWatchContextMenuRequested(const QPoint& pos)
...
+      QMenu* copyAddrSubmenu = contextMenu->addMenu(tr("Copy add&ress..."));
+      QAction* copyPointer = new QAction(tr("Copy &base address..."), this);
+      connect(copyPointer, &QAction::triggered, this,
+              [this, entry] { copyAddressToClipboard(entry->getConsoleAddress()); });
+      copyAddrSubmenu->addAction(copyPointer);
+      for (int i = 0; i < static_cast<int>(entry->getPointerLevel()); ++i)
+      {
+        std::string strAddressOfPath = entry->getAddressStringForPointerLevel(i + 1);
+        if (strAddressOfPath == "???")
+          break;
```
cristian64, line 290: Instead of testing strings, which is not very robust, could change in the future, and technically slower, why not verify the address directly?
```cpp
if (!entry->getAddressForPointerLevel(level))
    break;
```

Source/GUI/MemWatcher/MemWatchWidget.cpp
```diff
@@ -273,6 +273,43 @@ void MemWatchWidget::onMemWatchContextMenuRequested(const QPoint& pos)
...
+        QAction* showAddressOfPathInViewer = new QAction(
+            tr("Copy pointed address at &level %1...").arg(QString::number(i + 1)), this);
```
cristian64, line 292: Is `QString::number()` required? I think `QString::arg()` accepts numeric types as well as strings.

Source/GUI/MemWatcher/MemWatchWidget.cpp
```diff
@@ -393,6 +430,14 @@ void MemWatchWidget::pasteWatchFromClipBoard(const QModelIndex& referenceIndex)
   m_hasUnsavedChanges = true;
 }
 
+void MemWatchWidget::copyAddressToClipboard(u32 addr)
+{
+  char hex_string[8];
+  snprintf(hex_string, 8, "%X", addr);
+  QClipboard* clipboard = QApplication::clipboard();
+  clipboard->setText(hex_string);
```
cristian64, line 438: The use of the raw C array and `snprintf()` is error prone. Currently the array of size 8 could not possibly hold 8 characters _plus_ the null character (`\0`); garbage may be put in the clipboard.

Suggestion:
```cpp
const QString text{QString::number(addr, 16).toUpper()};
QApplication::clipboard()->setText(text);
```

Source/GUI/MemWatcher/MemWatchWidget.cpp
```diff
@@ -393,6 +430,14 @@ void MemWatchWidget::pasteWatchFromClipBoard(const QModelIndex& referenceIndex)
   m_hasUnsavedChanges = true;
 }
 
+void MemWatchWidget::copyAddressToClipboard(u32 addr)
```
cristian64, line 433: Nit: `addr` should be marked as `const`.

Source/GUI/MemWatcher/MemWatchWidget.h
```diff
@@ -39,6 +39,7 @@ class MemWatchWidget : public QWidget
   void copySelectedWatchesToClipBoard();
   void cutSelectedWatchesToClipBoard();
   void pasteWatchFromClipBoard(const QModelIndex& referenceIndex);
+  void copyAddressToClipboard(u32 addr);
```
cristian64, line 42: Technically, `copyAddressToClipboard()` could be `static`: it doesn't reference `this`. And, arguably, the function is general enough to be placed in a common module that could be reused later (although I don't have any suggestion).

Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
jahorta: I didn't realize that I could get the sender. This is definitely a better way.

Source/GUI/MemWatcher/MemWatchWidget.cpp
```diff
@@ -273,6 +273,43 @@ void MemWatchWidget::onMemWatchContextMenuRequested(const QPoint& pos)
...
+      QAction* copyPointer = new QAction(tr("Copy add&ress"), this);
+      connect(copyPointer, &QAction::triggered, this,
+              [this, entry] { copyAddressToClipboard(entry->getConsoleAddress()); });
+      contextMenu->addAction(copyPointer);
+
+      QModelIndexList selection = m_watchView->selectionModel()->selectedRows();
+      if (selection.count() == 0)
+      {
+        copyPointer->setEnabled(false);
+      }
```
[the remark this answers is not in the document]
jahorta: I realized that since it checks for a valid index to get to this scope, there is no way the selection would have zero entries, so I removed this check altogether

Source/GUI/MemWatcher/MemWatchWidget.cpp
jahorta: That makes sense. This is also simpler. With this I don't think I need a separate function to perform this, so I am going to integrate it into the context menu itself rather than be a separate method.

Source/GUI/MemWatcher/MemWatchWidget.cpp
jahorta: This method is no longer present, see above

Source/GUI/MemWatcher/MemWatchWidget.h
jahorta: This method is no longer present, see [above](https://github.com/aldelaro5/dolphin-memory-engine/pull/167#discussion_r1667768094)

dreamsyntax:
I'll be doing real-world use / tests of this feature branch later today. Is this otherwise ready for final review / merge? If it is, please mark it as non-draft with "Ready for review" button.

Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
```diff
@@ -424,3 +437,23 @@ MemWatchEntry* DlgAddWatchEntry::stealEntry()
   m_entry = nullptr;
   return entry;
 }
+
+void DlgAddWatchEntry::onPointerOffsetContextMenuRequested(const QPoint& pos)
+{
+  QLabel* const lbl = static_cast<QLabel*>(sender());
```
cristian64, line 443: Nit: In principle, it should use `qobject_cast`.

Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
```diff
@@ -424,3 +437,23 @@ MemWatchEntry* DlgAddWatchEntry::stealEntry()
...
+
+  const QString text{QString::number(lbl->property("addr").toUInt(), 16).toUpper()};
+  connect(copyAddr, &QAction::triggered, this,
+    [text] { QApplication::clipboard()->setText(text); });
+  contextMenu->addAction(copyAddr);
+
+  if (!lbl->property("addr").toUInt())
+    {
+        copyAddr->setEnabled(false);
+      }
```
cristian64, line 456: Nit: The indentation has gone bad around these lines (and perhaps in other lines too). The project uses Clang-Format for formatting; there is a `.clang-format` file in the repository.

One thing the build does in the CI is check that the formatting is correct. So, I'm guessing (and hoping!) that the CI would flag this as an error.

Source/GUI/MemWatcher/MemWatchWidget.cpp
```diff
@@ -273,6 +273,42 @@ void MemWatchWidget::onMemWatchContextMenuRequested(const QPoint& pos)
...
+      QMenu* const copyAddrSubmenu = contextMenu->addMenu(tr("Copy add&ress..."));
+      QAction* const copyPointer = new QAction(tr("Copy &base address..."), this);
+      const QString addrString{QString::number(entry->getConsoleAddress(), 16).toUpper()};
+      connect(copyPointer, &QAction::triggered, this,
+              [addrString] { QApplication::clipboard()->setText(addrString);
+      });
+      copyAddrSubmenu->addAction(copyPointer);
+      for (int i = 0; i < static_cast<int>(entry->getPointerLevel()); ++i)
+      {
+        std::string const strAddressOfPath = entry->getAddressStringForPointerLevel(i + 1);
```
cristian64, line 290: Is `strAddressOfPath` used?

Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
```diff
@@ -424,3 +437,23 @@ MemWatchEntry* DlgAddWatchEntry::stealEntry()
   m_entry = nullptr;
   return entry;
 }
+
+void DlgAddWatchEntry::onPointerOffsetContextMenuRequested(const QPoint& pos)
+{
+  QLabel* const lbl = static_cast<QLabel*>(sender());
+
+  QMenu* contextMenu = new QMenu(this);
```
cristian64, line 445: Nit: For correctness, both `contextMenu` (and `copyAddr` below) could be marked as `const`, so that the pointers cannot be reassigned inadvertently.

Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
```diff
@@ -424,3 +437,23 @@ MemWatchEntry* DlgAddWatchEntry::stealEntry()
...
+  QAction* copyAddr = new QAction(tr("&Copy Address"), this);
+
+  const QString text{QString::number(lbl->property("addr").toUInt(), 16).toUpper()};
```
cristian64, line 448: Question: Would we want to pad the string with heading 0s? (`14FA` vs `000014FA`)

Source/GUI/MemWatcher/MemWatchWidget.cpp
jahorta: It's not. I forgot to remove it when I changed the method for checking if a pointer address is valid. Removing it.

Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
jahorta: It wouldn't be necessary for MEM1. The "addr" property is set using the address value within the address space of Dolphin which has a base address of 0x80000000, so any offset into it should not need padding as it starts at 8 characters in hex. That being said, I'm not sure if MEM2 would need padding or not.

dreamsyntax:
Tested on Windows.

Works great, thanks. I definitely prefer this feature after having experimented with both, but there's no reason to remove the other context menu.

image

Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
dreamsyntax: Is there a blocker to merge this? @cristian64

dreamsyntax (approved):
As long as @cristian64 signs off on this, I'm good with merging as is.
Since we are not super strict about squash/merge rules, I'll leave it up to you if you want to squash @cristian64

cristian64:
Not tested, but looks good to me.

cristian64:
> As long as @cristian64 signs off on this, I'm good with merging as is. Since we are not super strict about squash/merge rules, I'll leave it up to you if you want to squash @cristian64

We should definitely squash this into a single commit.

PR #168, opened 2024-07-12
Upgrade Clang-Format check in GitHub Action.

cristian64:
The previous GitHub Action (`RafikFarhad/clang-format-github-action`) was not handling deeply nested files correctly: formatting was not enforced in a second directory level or deeper.

A more popular action (`jidicula/clang-format-action`) is now used, which enables Clang-Format 18.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,14 +10,14 @@ on:
 
 jobs:
   clang-format-check:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     name: Clang-Format Check
     steps:
       - uses: actions/checkout@v4
-      - uses: RafikFarhad/clang-format-github-action@v4
+      - uses: jidicula/clang-format-action@v4.13.0
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -10,14 +10,14 @@ on:
 
 jobs:
   clang-format-check:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     name: Clang-Format Check
     steps:
       - uses: actions/checkout@v4
-      - uses: RafikFarhad/clang-format-github-action@v4
+      - uses: jidicula/clang-format-action@v4.13.0
```
dreamsyntax, line 17: Is there not an official one you can use? Either way, looks good, thanks for catching this

dreamsyntax (approved):

.github/workflows/build.yml
cristian64: This is the more popular I could find. Happy to be proven wrong, though!

PR #174, opened 2024-08-18
Convert to `unsigned char` before invoking `std::isprint()`.

### LINKED ISSUE #173: Bug: Crash on Scrolling with Certain Watchlists
image

>	dolphin-memory-engine.exe!Common::formatMemoryToString(const char * memory, Common::MemType type, unsigned __int64 length, Common::MemBase base, bool isUnsigned, bool withBSwap) Line 654	C++

```cpp
      if (std::isprint(c))
```
in `MemoryCommon.cpp`

image

cristian64:
As the documentation states, the caller must check that the input character in [`std::isprint()`](https://en.cppreference.com/w/cpp/string/byte/isprint) is a valid `unsigned char`:

> Like all other functions from `<cctype>`, the behavior of `std::isprint` is undefined if the argument's value is neither representable as `unsigned char` nor equal to `EOF`. To use these functions safely with plain `char`s (or `signed char`s), the argument should first be converted to `unsigned char`:
>
> ```cpp
> bool my_isprint(char ch)
> {
>     return std::isprint(static_cast<unsigned char>(ch));
> }
> ```

The aforementioned undefined behavior manifests as a debug assertion when compiled with MSVC:

image

The affected functionality was introduced in #122.

Fixes #173.

### DIFF
--- a/Source/Common/MemoryCommon.cpp
+++ b/Source/Common/MemoryCommon.cpp
@@ -651,7 +651,7 @@ std::string formatMemoryToString(const char* memory, const MemType type, const s
       if (c == '\0')
         break;
 
-      if (std::isprint(c))
+      if (std::isprint(static_cast<int>(static_cast<unsigned char>(c))))
       {
         text.push_back(c);
       }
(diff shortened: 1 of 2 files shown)


dreamsyntax:
Thanks, beat me to it

dreamsyntax (approved):
Tested reproducing the crash on VC++, no longer occurs.

PR #175, opened 2024-08-23
Fix two crash bugs, add new context menu items

camila314:
- Fixed a bug where opening a context menu with nothing selected crashed the program
- Fixed a bug where adding pointer levels could crash the program
- Created new 'Add Watch' and 'Add Group' context menu items when not selecting anything

### DIFF
--- a/Source/GUI/MemWatcher/MemWatchWidget.cpp
+++ b/Source/GUI/MemWatcher/MemWatchWidget.cpp
@@ -267,6 +267,17 @@ void MemWatchWidget::onMemWatchContextMenuRequested(const QPoint& pos)
     contextMenu->addSeparator();
   }
 
+  MemWatchEntry* const entry{MemWatchModel::getEntryFromIndex(index)};
+  if (!entry) {
+    QAction* const addGroup{new QAction(tr("Add &group"), this)};
+    connect(addGroup, &QAction::triggered, this, &MemWatchWidget::onAddGroup);
+    contextMenu->addAction(addGroup);
+    QAction* const addWatch{new QAction(tr("Add &watch"), this)};
+    connect(addWatch, &QAction::triggered, this, &MemWatchWidget::onAddWatchEntry);
+    contextMenu->addAction(addWatch);
+    contextMenu->addSeparator();
+  }
+
   QAction* cut = new QAction(tr("Cu&t"), this);
   connect(cut, &QAction::triggered, this, [this] { cutSelectedWatchesToClipBoard(); });
   contextMenu->addAction(cut);
@@ -274,7 +285,6 @@ void MemWatchWidget::onMemWatchContextMenuRequested(const QPoint& pos)
   connect(copy, &QAction::triggered, this, [this] { copySelectedWatchesToClipBoard(); });
   contextMenu->addAction(copy);
 
-  MemWatchEntry* const entry{MemWatchModel::getEntryFromIndex(index)};
   if (entry)
   {
     if (entry->isBoundToPointer())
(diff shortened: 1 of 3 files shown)


Source/GUI/MemWatcher/MemWatchWidget.cpp
```diff
@@ -267,14 +267,24 @@ void MemWatchWidget::onMemWatchContextMenuRequested(const QPoint& pos)
     contextMenu->addSeparator();
   }
 
+  MemWatchEntry* const entry{MemWatchModel::getEntryFromIndex(index)};
+  if (!entry) {
```
cristian64, line 271: It kind of makes sense to me to also show these two actions if the right-clicked entry is a group. Thoughts?

cristian64 (commented):
The first of the crashes was introduced [by me] in https://github.com/aldelaro5/dolphin-memory-engine/pull/171. The stack trace is:
```
#0  MemWatchTreeNode::getEntry() const (this=0x0) at /w/dolphin-memory-engine/Source/MemoryWatch/MemWatchTreeNode.cpp:61
#1  0x000055b08f3e877d in MemWatchModel::getEntryFromIndex(QModelIndex const&) (index=...) at /w/dolphin-memory-engine/Source/GUI/MemWatcher/MemWatchModel.cpp:122
#2  0x000055b08f3f6c70 in MemWatchWidget::onMemWatchContextMenuRequested(QPoint const&) (this=0x55b091538000, pos=...) at /w/dolphin-memory-engine/Source/GUI/MemWatcher/MemWatchWidget.cpp:277
#3  0x00007f6377b6b023 in  () at /lib/x86_64-linux-gnu/libQt6Core.so.6
#4  0x00007f6378836889 in QWidget::customContextMenuRequested(QPoint const&) () at /lib/x86_64-linux-gnu/libQt6Widgets.so.6
```

I have not been able to reproduce the second of the crashes.

On a side comment, it would have been nice to keep the crash fixes in a PR, and the other enhancement on its own PR.

Source/GUI/MemWatcher/MemWatchWidget.cpp
camila314: probably but then one would need to code it so that the newly added watch-goes inside the clicked-on group

cristian64: I thought that was the case already. But if it isn't, then it can be looked at at another time.

Source/GUI/MemWatcher/MemWatchWidget.cpp
```diff
@@ -267,14 +267,24 @@ void MemWatchWidget::onMemWatchContextMenuRequested(const QPoint& pos)
     contextMenu->addSeparator();
   }
 
+  MemWatchEntry* const entry{MemWatchModel::getEntryFromIndex(index)};
+  if (!entry || node->isGroup()) {
```
cristian64, line 271: It'd probably be a bit more sensible to test `node` and not mix it in `entry`:
```cpp
if (!node || node->isGroup()) {
```
And the other line could be reverted back to its original place.

Source/GUI/MemWatcher/MemWatchWidget.cpp
```diff
@@ -267,6 +267,16 @@ void MemWatchWidget::onMemWatchContextMenuRequested(const QPoint& pos)
     contextMenu->addSeparator();
   }
 
+  if (!node || node->isGroup()) {
+    QAction* const addGroup{new QAction(tr("Add &group"), this)};
```
cristian64, line 271: It seems `G` is already used for the other **Group** action.

cristian64 (approved):

Issue #121, opened 2024-04-26
Feature: Autoload from Path


dreamsyntax:
While the automatic saving/loading of the watchlist from a temp file in appdata is useful, I find myself losing work in a multi user dmw if I forget to save.

- allow auto load/save on close to use a last opened/saved file (only when a persistant setting is set)
- prompt on closing? If temp file is being used.

cristian64:
As of #135, users will at least be warned, and will have a chance to save the file on exit if it has unsaved changes.

dreamsyntax:
#135 resolves issues with multiple DME instances - now merged.

Attached in a below use case I want to address before we can close this issue.

1. shared git repo with DMWs in it
2. open the DMW directly with DME, make changes. Save and quit. Commit changes. A day goes by, other people are working on the same DMW and add updates.
3. pull from git repo as someone updated it.
4. Later open DME, and since it auto keeps the changes from last DMW load, I forget to re-load from the file in the git repo (thus losing the changes from the other user's commits)

Effectively, I'm looking for an auto read from specific file directory mode rather than saving to/from the dolphin-memory-engine.ini.

cristian64:
In some offline discussion, these actions can be added to the **File** menu:
```
File > [x] Autoload Last File
     > [x] Autosave Last File
```
> C: Second option only enabled if first option is enabled, perhaps?
> D: I like it!
> C: If only first option is enabled, the prompt is shown.
> D: Yeah, that sounds ideal.

dreamsyntax:
Finally getting around to implementing this as it's become a pain not having it in my workflow.

![Image](https://github.com/user-attachments/assets/7ba1166c-a881-451e-ab42-be30bb9c819a)

I'm not sure what it is, but the text looks off...
Open to opinions on `Autosave` vs `Auto-save` vs `Auto Save`

dreamsyntax:
Nearing completion of Autoload.
After initial usage I'm thinking Autosave is a bit too risky/annoying without https://github.com/aldelaro5/dolphin-memory-engine/issues/182 implemented first.
Will release Autoload by itself, with closing prompting the existing unsaved changes, I think it may be overkill anyway.

dreamsyntax:
I'm definitely thinking Auto-save is overkill and somewhat dangerous, so unless someone wants that in the future, I won't make a new issue for it.

dreamsyntax:
Released in v1.2.5

PR #176, opened 2024-08-25
Retain file information when closing and reopening

camila314:
Previously, closing and reopening the program meant having to re-save the opened file and overwrite the file. This retains the opened file so you no longer have to go through that process.

Also changed "Clear the watch list" to "Reset watch list" and made it discard the saved file information.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ Source/.vscode/
 .vscode/
 .idea/*
 .DS_Store
+.cache
--- a/Source/GUI/MemWatcher/MemWatchWidget.cpp
+++ b/Source/GUI/MemWatcher/MemWatchWidget.cpp
@@ -267,7 +267,7 @@ void MemWatchWidget::onMemWatchContextMenuRequested(const QPoint& pos)
     contextMenu->addSeparator();
   }
 
-  if (!node || node->isGroup())
+  if (node == m_watchModel->getRootNode() || node->isGroup())
   {
     QAction* const addGroup{new QAction(tr("Add gro&up"), this)};
     connect(addGroup, &QAction::triggered, this, &MemWatchWidget::onAddGroup);
--- a/Source/GUI/MemWatcher/MemWatchWidget.cpp
+++ b/Source/GUI/MemWatcher/MemWatchWidget.cpp
@@ -853,6 +853,7 @@ void MemWatchWidget::clearWatchList()
     return;
 
   m_watchModel->clearRoot();
+  m_watchListFile = "";
 }
 
 void MemWatchWidget::importFromCTFile()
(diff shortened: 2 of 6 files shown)


dreamsyntax:
I like the intent behind this, but I still would prefer something more like this: https://github.com/aldelaro5/dolphin-memory-engine/issues/121
Your change is basically what I want, but I also wanted a way to toggle the modes, incase working in a scratch space.

The clear/reset change sounds good.

camila314:
the best system would be to make it document-based (like Bit Slicer for macOS) but i fear i don't know enough about qt to make that change, maybe some day though

dreamsyntax:
Just to be clear, so we don't break the flow of existing users I would want
https://github.com/aldelaro5/dolphin-memory-engine/issues/121#issuecomment-2106374792

```
File > [x] Autoload Last File
     > [x] Autosave Last File
```

Something like that in the PR that enables this behavior.
It could just be one toggle.
File -> [x] Auto Load/Save File

dreamsyntax:
Bumping.

@cristian64 thoughts on just merging as is? I like the general idea even if I prefer optionally maintaining the old way.

Source/GUI/MemWatcher/MemWatchWidget.cpp
```diff
@@ -267,7 +267,7 @@ void MemWatchWidget::onMemWatchContextMenuRequested(const QPoint& pos)
     contextMenu->addSeparator();
   }
 
-  if (!node || node->isGroup())
+  if (node == m_watchModel->getRootNode() || node->isGroup())
```
cristian64, line 270: _Ideally_, this change (which relates to https://github.com/aldelaro5/dolphin-memory-engine/pull/175) should land in its own PR.

Source/GUI/MemWatcher/MemWatchWidget.cpp
```diff
@@ -853,6 +853,7 @@ void MemWatchWidget::clearWatchList()
     return;
 
   m_watchModel->clearRoot();
+  m_watchListFile = "";
```
cristian64, line 856: To avoid constructing another object for the sake of clearing the member:

```diff
- m_watchListFile = "";
+ m_watchListFile.clear();
```

.gitignore
```diff
@@ -17,4 +17,5 @@ Source/.vs/
 Source/.vscode/
 .vscode/
 .idea/*
-.DS_Store
\ No newline at end of file
+.DS_Store
+.cache
```
cristian64, line 21: Out of curiosity, what is `.cache`?

And, as always, these unrelated changes should at the very least have their own commit and description. Perhaps it can even have a dedicated PR to avoid getting the change blocked by other potentially controversial changes.

cristian64:
> Bumping.
> 
> @cristian64 thoughts on just merging as is? I like the general idea even if I prefer optionally maintaining the old way.

I may be a bit out of the loop, but I'm not entirely sure what these are trying to accomplish. By looking at the diff, I can't see how the watch list file path is consumed on startup.

dreamsyntax:
@camila314 following up if you plan to make changes, I'm going to release another patch version soon

PR #180, opened 2025-01-20
Cleanup: Remove MSBuild / SLN

dreamsyntax:
* Update README and gitignore
* Commit custom debug build paths for CMake VS

Added some paths for gitignore that I think would be common to accidentally generate if you are new to CMake on Windows and using VS or vscode(ium).

Now that Dolphin is in talks to drop the .sln, I think we may as well get ahead of it.

I did some testing with CMake in Visual Studio, it works fine for debugging so I'm satisfied.
The only thing is release variants are not set up currently, you can only build the debug variant locally.


Source/CMakeSettings.json
```diff
@@ -0,0 +1,15 @@
+﻿{
```
cristian64, line 1: My question is whether keeping this one version in the project is a sensible thing to do. Normally I'd prefer these to stay local to the dev's machine.

dreamsyntax: I'm fine removing it and adding to gitignore.
The reason I added it is the default is /out rather than matching msbuild /bin.

cristian64: My preference would be to add the entry to the ignore file, but obviously I don't have a strong opinion and it doesn't really affect me. If you think others will find it useful, go ahead.

cristian64:
Happy to see this landing. Then #89 can be closed.

Source/CMakeSettings.json
dreamsyntax: I'll remove it for now since I'm not super familiar with CMake on Windows.
It didn't really *add/make anything easier* by changing the /out to /bin.
Ultimately it doesn't matter, so removing it.

Issue #179, opened 2025-01-16
"Binary" variable type addition


Jaohammed:
Hi there, it would be really useful if we could add variables that had the "binary" type, as some games have bitfields where each bit represents an individual thing, and as of now it's not possible to make each bit in the field its own individual variable for proper labelling.

Below is an example of what can be done using Cheat Engine, but can't be done on DME as of now.

![image](https://github.com/user-attachments/assets/45fdacaf-df45-4b07-8f3d-3b54f96c09b5)

Lastly, as a long time user of DME let me thank you for making this software, it has been an invaluable tool for TASes! Even if this feature can't be added, everything we already have is plenty, this is just to make labelling a very specific type of data possible.

dreamsyntax:
Could you show an example with an actively attached game? Im not sure I follow how this is different from changing the View mode to Binary on an array in DME

Jaohammed:
Yeah, I can already see the values just fine and edit them individually, it's just a matter of not being able to assign labels for each of the bits individually, let me attach a screenshot below

![image](https://github.com/user-attachments/assets/d5f7153f-975b-44d4-8a01-fa5d5ef74151)

The highlighted fields there are 2 halves of a bitfield where each bit corresponds to an upgrade for the character I'm currently playing as, and it would be really useful to be able to have labels for each upgrade, because as of now we have to either know them by memory for consult a table, or just try values and see what upgrade we get.

The same thing applies to the other bitfields, but the upgrades one is the only one we have the exact information about each bit so far.

dreamsyntax:
@cristian64 do you think this would require a node watch node type to accomplish?

Jaohammed:
I don't know if this is helpful, and you probably already know it, but here's how the variable is set up in Cheat Engine
![image](https://github.com/user-attachments/assets/1eee7098-a842-443e-867a-5b90c28421c0)

cristian64:
I think the request makes sense.

> [@cristian64](https://github.com/cristian64) do you think this would require a node watch node type to accomplish?

It almost feels like this type would be similar to an array with some extra metadata (bit offset and bit count), so perhaps it could be an extension of that with an alternative representation in the widget.

PR #181, opened 2025-02-09
Adding a struct type with editor

jahorta:
This is an attempt to introduce a new struct type. This type would allow the user to define a block of memory using a list of entries with any type. Then the struct would automatically be created when assigned to a memory watch. This includes a number of new features to accomplish this:

- A struct definition class (StructDef) which holds the definition of a struct.
	- This contains a name, length, and list of fields

- A struct field definition class (FieldDef) which holds the field information.
	- The offset within the struct. These are generated based on the size of the fields before it. 
	- the field's size within the struct
	- a MemWatchEntry defining the type of the field.
		- All current types are supported, including structs and pointers to structs. For these last two, the inter-struct references are stored in case of name or field changes.
		- The only restrictions are that there can be no cyclic references to structs as it would make it impossible to determine the size of the struct. Pointers to structs do not have this restriction.
		- When editing the entry of a struct field, The DlgAddWatchEntry class ignores the address and preview.
	- Struct fields are currently checked for validity by making sure they do not overlap each other.

- A struct editor widget as a separate window to organize and edit struct definitions.
	- Tree-style organization to allow grouping of structs.
	- Struct definitions are identified using a "Namespace" created from combining group names and the struct name.
	- Struct definitions can be edited using the context menu. When editing a struct, its fields will appear in a tableview on the right. Since struct offsets are currently generated from the field definitions, the user can enter padding fields to get a field to the correct offset.
	- When editing a struct, a copy is made to edit. If a change has been made to a struct, the save button will be enabled. The state of this button is used to determine whether there are any unsaved changes on the current struct.

- The ability to select a struct in the DlgAddWatchEntry for memory watches.
	- A list of names is currently passed to the DlgAddWatchEntry to select from and a simple combo box is shown to the user to select a struct from.

- The automatic creation and deletion of memory watches using struct definitions as a template.
	- When a mem watch node has a struct type entry, if its definition is known it will get a placeholder watch node as a child. 
	- If the node is expanded, the placeholder node will be deleted and the fields of the struct definition will be copied as new child nodes under the struct node.
	- If a node is collapsed, the child nodes will be removed and a placeholder node will be added again.

- Saving and loading the struct tree to/from the settings on application start and close.

- Saving and loading ~~any currently used~~ all struct definitions to/from a memory watch file.
	- When loading from a file, any namespace collisions are presented to the user to resolve. If a new name is needed, it currently just auto-generates one.

This is a draft pull request while I do some testing, bug fixing, and refactoring, but has most of the features fully implemented, so I wanted some feedback. I am open to any feedback, but am wondering about a few design choices in particular:

- In the mem watcher, structs are expandable, so if a struct type with a known struct definition is found it will be given a "placeholder" child to allow the user to expand the struct. At the moment, the fields of the struct will be created on expansion and deleted on collapsing of the struct. 
	1. I think the creation on expansion is necessary because the struct can handle pointers to its own type and in a situation where you have a doubly linked list you would have infinite generation if you try to generate everything since the next and previous pointers would form cycles.
	2. I am not sure if the deletion of the field entries on collapse of a node is necessary, or even good. The reason for this was that the hidden entries would take up memory and processing power on update that is not visible to the user which may eventually lead to problems, but the amount of memory that is taken by each entry is so small that I don't really see it being a problem. Also, since fields of a struct are deleted at the moment, expansion of any struct fields is lost, so the user would need to re-expand any nodes that they had open previously.

- I added an icon to identify structs in the mem watcher. 
	1. Is this easy to see?
	2. Should there also be an icon for fields of a struct since they are not as flexible as a normal watch entry?

- Struct fields are currently checked to make sure that they do not overlap. This is not strictly necessary, and there is a case to be made for fields that may overlap (Unions, bitfields, etc.).
 	1. Should the user be allowed to overlap struct fields? 
	2. Should there be a checkbox for this?


dreamsyntax:
This feature sounds great. Ill be trying some builds of this in the interim. Review is huge so it may be a while, even after you've fully completed it.

dreamsyntax:
Hey there, I know this is WIP so I'm not expecting action on this, but when its closer to being finished, the commits will have to be sorted out:
image

jahorta:
Definitely, I'll rebase the commits into a single commit before review.

cristian64:
> Definitely, I'll rebase the commits into a single commit before review.

...but notice that a single commit would be unwieldy from a code review point of view.

jahorta:
Okay. I'm still pretty new to pull requests, how should I organize the commits so that they are easier to review?

dreamsyntax:
For now just focus on your feature and when its done we can rediscuss. Its already an unmanageable amount of commits, so easiest scenario is squashing post completion

jahorta:
So I think I have a fairly stable branch now. I would like to include the recent commits to the aldelaro5:master branch to see if my changes play nicely with them and see if I need to make any more changes to my branch. What would be the best way to do this? a merge or a rebase with the master branch?

dreamsyntax:
I would highly recommend while on this branch, you checkout to a new branch and push it as a backup before trying to merge or rebase.

`git checkout -b backup-struct-feature`
`git push` (you may need -u if you dont have auto remote enabled, it will tell you with the command)
`git checkout -` (back to this branch)

I stress to ensure you have a copy via the above to avoid losing work.

jahorta:
> I would highly recommend while on this branch, you checkout to a new branch and push it as a backup before trying to merge or rebase.

Yeah, that makes sense. Okay, I did that.

dreamsyntax:
I'm attempting a rebase myself (on another branch) - Regarding:
`SConfig::getInstance().setStructDefs(m_structEditor->saveStructTree());`

Is this something that should be saved to the temporary store OR a dmw file, or both?

jahorta:
I currently have it set up to save the struct definition tree to the settings file when DME is closed, and that is the main source of all defined structs. At the moment, when a dmw file is saved I only include those structs that are used in the saved watch list, so it will only be a subset. I would say we should load struct defs from the settings, then if a watch list is auto-loaded, load those struct definitions on top. That way the user has full access to all structs that they have defined. When loading the dmw file, the struct editor is able to detect conflicts like reuse of a struct name or changes in a struct definition and should ask for direction from the user on how to resolve them. This can also be changed if there is a better way to handle it.

dreamsyntax:
> I currently have it set up to save the struct definition tree to the settings file when DME is closed, and that is the main source of all defined structs. At the moment, when a dmw file is saved I only include those structs that are used in the saved watch list, so it will only be a subset. I would say we should load struct defs from the settings, then if a watch list is auto-loaded, load those struct definitions on top. That way the user has full access to all structs that they have defined. When loading the dmw file, the struct editor is able to detect conflicts like reuse of a struct name or changes in a struct definition and should ask for direction from the user on how to resolve them. This can also be changed if there is a better way to handle it.

That seems fine to me.
I rebased with the assumption that even if autoloading a file, you would load from the shared struct definitions.
Wouldn't this have conflicts if viewing different games though? Maybe I'll understand better once I try your build out.

Artifacts will be generated here, which it rebased on master:
https://github.com/aldelaro5/dolphin-memory-engine/pull/181

dreamsyntax:
Initial inspection:
 * Right clicking and clicking "add field" to a new struct crashes DME
 * Right clicking in the watchlist crashes DME

jahorta:
> Initial inspection:
> 
> * Right clicking and clicking "add field" to a new struct crashes DME
> * Right clicking in the watchlist crashes DME

Would you like me to fix the bug in the structs-test-rebased branch, or rebase here and do it here?

jahorta:
> > I currently have it set up to save the struct definition tree to the settings file when DME is closed, and that is the main source of all defined structs. At the moment, when a dmw file is saved I only include those structs that are used in the saved watch list, so it will only be a subset. I would say we should load struct defs from the settings, then if a watch list is auto-loaded, load those struct definitions on top. That way the user has full access to all structs that they have defined. When loading the dmw file, the struct editor is able to detect conflicts like reuse of a struct name or changes in a struct definition and should ask for direction from the user on how to resolve them. This can also be changed if there is a better way to handle it.
> 
> That seems fine to me. I rebased with the assumption that even if autoloading a file, you would load from the shared struct definitions. Wouldn't this have conflicts if viewing different games though? Maybe I'll understand better once I try your build out.

Honestly, I never considered multiple games since I only ever work with one... The main reason I put them in the settings was so that they were always available if you wanted to work on a different dmw file for the same game. I have a number of different dmw files for investigating different systems in the game, so that seemed to make sense at the time. How about this:

To account for multiple games, it might be better to save all structs in the struct editor to a dmw file, not just the ones in use by MemWatchEntries.
Then in the dialog asking to merge watch lists, add a button that says "Structs Only" so there's an option to just add any struct definitions from the incoming file. 
Otherwise, if the watchlists are merged (Yes button), add any structs from the file also (with checking for overlap/definition changes). 
If the watchlists are not merged (No button), delete any structs present in the struct editor and load any present in the dmw file.
It would still save the struct tree to the settings when DME is closed, but it will only load structs from the settings if the user wants to restore the watch model.

dreamsyntax:
For now just use your current working branch. Don't worry about the rebase ones. I was just using it to experiment.

jahorta:
Okay, so I did some bug fixing and also reworked the save/load structs to work as I posted before:

> To account for multiple games, it might be better to save all structs in the struct editor to a dmw file, not just the ones in use by MemWatchEntries.
> Then in the dialog asking to merge watch lists, add a button that says "Structs Only" so there's an option to just add any struct definitions from the incoming file.
> Otherwise, if the watchlists are merged (Yes button), add any structs from the file also (with checking for overlap/definition changes).
> If the watchlists are not merged (No button), delete any structs present in the struct editor and load any present in the dmw file.
> It would still save the struct tree to the settings when DME is closed, but it will only load structs from the settings if the user wants to restore the watch model.

I also added some convenience features: 
- drag/drop reordering or copying (hold control while dropping, it uses Qt::ControlModifier) of struct fields in the detail editor
- duplication of structs (only duplicates single structs, will not duplicate if multiple tree nodes are selected)
- delete key now triggers delete method for both the select widget (deleting nodes) and the details widget (deleting fields)

Now I'm planning to spend some time looking for bugs and any major pain points that might need to be fixed.

dreamsyntax:
Is this in a good state to start experimenting again on my end?
I noticed you reduced it to 43 commits now, is this rebased?

jahorta:
Yeah, I did some squashing and it is now rebased on all the commits that are in the current aldelaro5:master. I fixed the bugs you found, plus a number of other ones, so I think it is good to start experimenting. 
There's still a visual bug I need to find and fix (when you have a struct watch that is expanded but you edit and save that struct definition in the editor, it will appear like the struct was deleted in the watch, but it can be fixed by collapsing and re expanding that watch with that struct type). I haven't found any other crashing bugs yet.

Source/GUI/MainWindow.cpp
```diff
@@ -719,3 +748,9 @@ void MainWindow::updateStatusBar()
   const QString toolTip{toolTipLines.join("\n\n")};
   m_statusLabel->parentWidget()->setToolTip(toolTip);
 }
+
+void MainWindow::makeStructEditor()
+{
+  m_structEditor = new StructEditorWidget(nullptr);
```
dreamsyntax, line 754: why nullptr?

Source/GUI/GUICommon.h
```diff
@@ -13,6 +13,7 @@ extern QStringList g_memBaseNames;
 
 QString getStringFromType(Common::MemType type, size_t length = 0);
 QString getNameFromBase(Common::MemBase base);
+bool isContainerType(Common::MemType type);
```
dreamsyntax, line 16: Naming, do you intend Container vs Struct? Are there plans to expand using Containers as a base for features like bitfields? https://github.com/aldelaro5/dolphin-memory-engine/issues/179

Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
```diff
@@ -16,9 +16,13 @@
...
 
 DlgAddWatchEntry::DlgAddWatchEntry(const bool newEntry, MemWatchEntry* const entry,
-                                   QWidget* const parent)
+                                   QVector<QString> const structs, QWidget* const parent,
+                                   bool isForStructField)
     : QDialog(parent)
 {
+  m_isForStructField = isForStructField;
+  m_structNames = structs;
+  m_structNames.push_front(QString(""));
```
dreamsyntax, line 25:
```diff
- m_structNames.push_front(QString(""));
+ m_structNames.push_front(QString());

File: Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp [hunk 4]
```diff
@@ -319,8 +368,19 @@ void DlgAddWatchEntry::accept()
...
+    QString errorMsg =
+        tr("A struct name must be selected with the struct type, it cannot be an empty string");
+    QMessageBox* errorBox = new QMessageBox(QMessageBox::Critical, tr("Invalid Struct Type"),
+                                            errorMsg, QMessageBox::Ok, this);
+    errorBox->exec();
+  }
   else
   {
+    if (m_isForStructField)
+      m_entry->setConsoleAddress(0);
```
dreamsyntax, line 383: note for self - inspect

Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
```diff
@@ -349,6 +409,10 @@ void DlgAddWatchEntry::accept()
       m_entry->setLabel(m_txbLabel->text());
     m_entry->setBase(Common::MemBase::base_decimal);
     m_entry->setSignedUnsigned(false);
+    if (m_entry->getType() == Common::MemType::type_struct)
+      m_entry->setStructName(m_structNames[m_structSelect->currentIndex()]);
+    else
+      m_entry->setStructName(QString(""));
```
dreamsyntax, line 415:
```diff
-m_entry->setStructName(QString(""));
+m_entry->setStructName(QString());
```

Source/GUI/MemWatcher/Dialogs/DlgChangeType.cpp
```diff
@@ -8,12 +8,17 @@
 
 #include "../../GUICommon.h"
 
-DlgChangeType::DlgChangeType(QWidget* parent, const int typeIndex, const size_t length)
-    : QDialog(parent), m_typeIndex(typeIndex), m_length(length)
+DlgChangeType::DlgChangeType(QWidget* parent, const int typeIndex, const size_t length,
+                             QVector<QString> structNames, QString curStructName)
+    : QDialog(parent), m_typeIndex(typeIndex), m_length(length), m_structNames(structNames)
 {
+  m_structNames.push_front(QString(""));
```
dreamsyntax, line 15:
```diff
-m_structNames.push_front(QString(""));
+m_structNames.push_front(QString());
```

Source/GUI/MemWatcher/MemWatchModel.cpp
```diff
@@ -416,16 +503,21 @@ Qt::ItemFlags MemWatchModel::flags(const QModelIndex& index) const
...
-    flags |= Qt::ItemIsDropEnabled;
+    flags |= Qt::ItemIsDropEnabled | Qt::ItemIsDragEnabled;
     if (index.column() == WATCH_COL_LABEL)
       flags |= Qt::ItemIsEditable;
     return flags;
   }
+  else if (node->getEntry()->getType() == Common::MemType::type_none)
+    return Qt::ItemFlag::NoItemFlags;
+  else if (parent == m_rootNode || parent->isGroup())
+    flags |= Qt::ItemIsDragEnabled;
```
dreamsyntax, line 520: Functionality change: Misses line 515 originally having ItemIsDropEnabled due to early return

dreamsyntax (changes requested):

Source/GUI/MainWindow.cpp
jahorta: That argument is for the parent widget. If I pass in a reference to the main window, the struct editor will not show up as a separate window, so I passed in a nullptr. I based it off of the makeMemViewer code which also makes a new window.
```
void MainWindow::makeMemViewer()
{
  m_viewer = new MemViewerWidget(nullptr);
  m_viewer->setWindowIcon(windowIcon());
  connect(m_viewer, &MemViewerWidget::mustUnhook, this, &MainWindow::onUnhook);
  connect(m_viewer, &MemViewerWidget::addWatchRequested, m_watcher, &MemWatchWidget::addWatchEntry);
  connect(m_watcher, &MemWatchWidget::goToAddressInViewer, this,
          &MainWindow::onOpenMemViewerWithAddress);
}
```

Source/GUI/GUICommon.h
dreamsyntax: Sounds reasonable to me. Thanks for confirming.

Source/GUI/StructEditor/StructDetailModel.h
```diff
@@ -0,0 +1,80 @@
...
+    STRUCT_COL_TYPE,
+    STRUCT_COL_LABEL,
+    STRUCT_COL_NUM
+  };
+
+  explicit StructDetailModel(QObject* parent);
+  ~StructDetailModel() override;
+
+  StructDetailModel(const StructDetailModel&) = delete;
+  StructDetailModel(StructDetailModel&&) = delete;
```
dreamsyntax, line 26: self note for later

Source/GUI/StructEditor/StructDetailModel.h
```diff
@@ -0,0 +1,80 @@
...
+  void modifyStructReference(QString nodeName, QString target, bool addIt, bool& ok);
+  void modifyStructPointerReference(QString nodeName, QString target, bool addIt);
+
+private:
+  QString getFieldDetails(FieldDef* field) const;
+  void removeFields(int start, int count = 1);
+  void updateFieldOffsets();
+  void reduceIndicesToRows(QModelIndexList& indices);
+
+  StructTreeNode* m_baseNode = nullptr;
```
dreamsyntax, line 77: self note for later

Source/MemoryWatch/MemWatchEntry.cpp
```diff
@@ -29,13 +32,15 @@ MemWatchEntry::MemWatchEntry()
   *m_memory = 0;
   m_isUnsigned = false;
   m_consoleAddress = 0x80000000;
+  m_curActualAddress = 0x80000000;
```
dreamsyntax, line 35: self note for later

Source/Resources/struct.svg
dreamsyntax, line 1: Just double checking on these, what is the source of these two SVGs

Source/Structs/FieldDef.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+void FieldDef::setFieldSize(u32 size)
+{
+  m_size = size;
+}
+
+QString FieldDef::getLabel() const
+{
+  if (!m_entry)
+    return QString("");
```
dreamsyntax, line 82:
```diff
- return QString("");
+ return QString();
```

Source/Structs/FieldDef.cpp
```diff
@@ -0,0 +1,246 @@
...
+    descLines += QString("Struct Name: %1").arg(m_entry->getStructName());
+  size_t size = Common::getSizeForType(m_entry->getType(), m_entry->getLength());
+  descLines += QString("Entry Length: %1").arg(size);
+  descLines += QString("Is Pointer: %1").arg(m_entry->isBoundToPointer() ? "Yes" : "No");
+  if (m_entry->isBoundToPointer())
+  {
+    int i = 0;
+    while (i < static_cast<int>(m_entry->getPointerLevel()))
+    {
+      descLines += QString("  Pointer Offset %1: %2").arg(i).arg(m_entry->getPointerOffset(i));
```
dreamsyntax, line 217: Is the double space intentional? I assume so, will look for this in user testing or if you have a screenshot.

Source/Structs/StructDef.cpp
```diff
@@ -0,0 +1,262 @@
+#include <QJsonArray>
+
+#include "StructDef.h"
+
+StructDef::StructDef()
+{
+  m_label = QString("");
```
dreamsyntax, line 7:
```diff
-m_label = QString("");
+m_label = QString();
```

Source/Structs/StructTreeNode.cpp
```diff
@@ -0,0 +1,336 @@
...
+    m_structDef->writeToJson(structDefJson);
+    json["struct"] = structDefJson;
+  }
+}
+
+QVector<QString> StructTreeNode::getStructNames(bool includeGroups, QString prefix)
+{
+  QVector<QString> names;
+
+  if (prefix == QString("") && m_parent)
```
dreamsyntax, line 246:
```diff
-if (prefix == QString("") && m_parent)
+if (prefix == QString() && m_parent)

File: Source/Structs/StructTreeNode.h [hunk 15]
```diff
@@ -0,0 +1,66 @@
...
+  void insertChild(int row, StructTreeNode* node);
+  void removeChild(int row);
+  void removeChild(StructTreeNode* child);
+  void removeChildren();
+  void deleteChildren();
+
+  void readFromJson(const QJsonObject& json, StructTreeNode* parent = nullptr);
+  void writeToJson(QJsonObject& json) const;
+
+  QVector<QString> getStructNames(bool includeGroups = false, QString prefix = QString(""));
```
dreamsyntax, line 49:
```diff
-QVector<QString> getStructNames(bool includeGroups = false, QString prefix = QString(""));
+QVector<QString> getStructNames(bool includeGroups = false, QString prefix = QString());

dreamsyntax (changes requested):

Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
jahorta: changed

Source/GUI/MemWatcher/Dialogs/DlgChangeType.cpp
jahorta: changed

Source/GUI/MemWatcher/Dialogs/DlgAddWatchEntry.cpp
jahorta: changed

Source/Structs/StructDef.cpp
jahorta: changed

Source/Structs/StructTreeNode.cpp
jahorta: changed

Source/Structs/StructTreeNode.h
jahorta: changed

Source/Structs/FieldDef.cpp
jahorta: image
The double space gives the indentation for the last line in the image.

Source/GUI/MemWatcher/MemWatchModel.cpp
dreamsyntax: You are right, I misread the lines.

jahorta:
> @jahorta Do you have a preference if we squash this as one commit on merge? Given the scope of this I would prefer it, but I'm open if you prefer having it broken apart.
> 
> I'm not expecting this to work perfectly yet. I'm fine with releasing this once merged as a pre-release for a while, with a warning to keep watchlists backed up incase bugs with the dmw modification are uncovered.

I have no preference with respect to how many commits it ends up as. If we can make this a pre-release, I have a few small changes I would like to make before we mark it ready for review.

jahorta:
> Just following up in if you have plans to continue this. No rush though, appreciate whenever you get to it.

Sorry about that. Had a busy month at work. I think its ready for review, so I'll change its status.

dreamsyntax (approved):
I'm thinking we release this as a pre-release at this point, the code LGTM and in my light usage its stable enough. Definitely will put warnings there may be issues with it though.

dreamsyntax:
Preview release created https://github.com/aldelaro5/dolphin-memory-engine/actions/runs/15030886845

Issue #182, opened 2025-02-22
Feature: Display 'active' opened .DMW file


dreamsyntax:
* Add the filename opened in the status bar
* When in 'scratch' state / using the stored watchlist should be indicated somehow
* Potentially just show the filename & last directory name, but absolute path is fine too.

![Image](https://github.com/user-attachments/assets/9bc05da2-1e7f-454f-9ab7-7d7dcadecefd)

example:

![Image](https://github.com/user-attachments/assets/9902aa7f-7d3f-4669-9bf5-ba6b6d52c159)

dreamsyntax:
Debating on the wording for the case of no file loaded:

![Image](https://github.com/user-attachments/assets/1f4ce86a-923d-47f1-85e8-efe76acc52da)

![Image](https://github.com/user-attachments/assets/7fc45675-4419-41ad-a193-bedcfd7e058d)

--
File loaded looks like (on Windows)

![Image](https://github.com/user-attachments/assets/01637469-8b2c-42e1-bac2-36db4fc95c78)

dreamsyntax:
Release in v1.2.5.

In case no file loaded, the file path will be omitted rather than adding text.

PR #183, opened 2025-02-22
Add Autoload Last File Option

dreamsyntax:
Implements https://github.com/aldelaro5/dolphin-memory-engine/issues/121

image

This feature autoloads the last .DMW you have previously loaded.
This is useful if working on a .DMW file you regularly commit to a shared repository.

- Enabling the feature will stop saving to the 'temporary' watchlist space.
- Disabling the feature will clear your current watchlist (giving you an opportunity to save it if you have not), and then load the old 'temporary' watchlist.
- Closing DME will now prompt you to save if this mode is on and you have not saved

---

## Changes
* Clearing the watchlist now also drops the file reference
* Closing DME while the Auto-load checkbox is enabled will prompt you if you have unsaved changes, just like if two DME instances are running.
* This is NOT enabled by default.
* Makes `m_watchListFile` public - necessary to get file location, and to be used later in https://github.com/aldelaro5/dolphin-memory-engine/issues/182 

<details><summary>TODOs (All Done)</summary>
[WONTDO]  // TODO: Potentially restoreWatchModel first, then openWatchFile in autoload scenario. If they don't match, warn/prompt unsaved changes in model.

[x] Do we want auto-load by default - _No, for now we will default to off_

[x] Potentially a way to switch between 'scratch' (the stored watchmodel in the config) and files on the fly? - _The config watchlist will not be saved, that way you can turn off the feature to switch between them_

[x] Lint

[x] Ensure when a user toggles the feature on/off the stored watchmodel is *not* overwritten. Warn the user if unsaved changes, and then reload the watchlist, or prompt to save it

[x] Don't save the config/scratch stored watchlist if this feature is enabled - prevents data loss

</details>

### DIFF
--- a/Source/GUI/MainWindow.cpp
+++ b/Source/GUI/MainWindow.cpp
@@ -41,7 +41,17 @@ MainWindow::MainWindow()
   GUICommon::changeApplicationStyle(
       static_cast<GUICommon::ApplicationStyle>(SConfig::getInstance().getTheme()));
 
-  m_watcher->restoreWatchModel(SConfig::getInstance().getWatchModel());
+  m_actAutoloadLastFile->setChecked(SConfig::getInstance().getAutoloadLastFile());
+
+  if (m_actAutoloadLastFile->isChecked() && !SConfig::getInstance().getLastLoadedFile().isEmpty())
+  {
+    m_watcher->openWatchFile(SConfig::getInstance().getLastLoadedFile());
+  }
+  else
+  {
+    m_watcher->restoreWatchModel(SConfig::getInstance().getWatchModel());
+  }
+
   m_actAutoHook->setChecked(SConfig::getInstance().getAutoHook());
 
   if (m_actAutoHook->isChecked())
@@ -66,6 +76,8 @@ void MainWindow::makeMenus()
   m_actClearWatchList = new QAction(tr("&Clear the watch list"), this);
   m_actImportFromCT = new QAction(tr("&Import from Cheat Engine's CT file..."), this);
   m_actExportAsCSV = new QAction(tr("&Export as CSV..."), this);
+  m_actAutoloadLastFile = new QAction(tr("Auto-load last file"), this);
+  m_actAutoloadLastFile->setCheckable(true);
   QAction* const actOpenConfigDir{new QAction(tr("Open Configuration Directory..."), this)};
 
   m_actOpenWatchList->setShortcut(Qt::Modifier::CTRL | Qt::Key::Key_O);
@@ -105,6 +117,9 @@ void MainWindow::makeMenus()
     QDesktopServices::openUrl(url);
   });
 
+  connect(m_actAutoloadLastFile, &QAction::triggered, this,
+          &MainWindow::onAutoLoadLastFileTriggered);
+
   connect(m_actSettings, &QAction::triggered, this, &MainWindow::onOpenSettings);
 
   connect(m_actAutoHook, &QAction::toggled, this, &MainWindow::onAutoHookToggled);
@@ -126,6 +141,8 @@ void MainWindow::makeMenus()
   m_menuFile->addSeparator();
   m_menuFile->addAction(m_actImportFromCT);
   m_menuFile->addAction(m_actExportAsCSV);
+  m_menuFile->addSeparator();
+  m_menuFile->addAction(m_actAutoloadLastFile);
   m_menuFile->addSeparator();
   m_menuFile->addAction(actOpenConfigDir);
   m_menuFile->addSeparator();
@@ -334,6 +351,16 @@ void MainWindow::onUnhook()
   updateDolphinHookingStatus();
 }
 
+void MainWindow::onAutoLoadLastFileTriggered(const bool checked)
+{
+  if (checked)
+    return;
+  if (!m_watcher->warnIfUnsavedChanges())
+    return;
+  m_watcher->clearWatchList();
+  m_watcher->restoreWatchModel(SConfig::getInstance().getWatchModel());
+}
+
 void MainWindow::onAutoHookToggled(const bool checked)
 {
   if (checked)
@@ -384,7 +411,7 @@ void MainWindow::onSplitterMoved(const int pos, const int index)
 void MainWindow::onOpenWatchFile()
 {
   if (m_watcher->warnIfUnsavedChanges())
-    m_watcher->openWatchFile();
+    m_watcher->openWatchFile(QString());
 }
 
 void MainWindow::onSaveWatchFile()
@@ -545,13 +572,19 @@ void MainWindow::onQuit()
 void MainWindow::closeEvent(QCloseEvent* event)
 {
   SConfig::getInstance().setAutoHook(m_actAutoHook->isChecked());
-  SConfig::getInstance().setWatchModel(m_watcher->saveWatchModel());
+  SConfig::getInstance().setAutoloadLastFile(m_actAutoloadLastFile->isChecked());
+  SConfig::getInstance().setLastLoadedFile(m_watcher->m_watchListFile);
+  if (!m_actAutoloadLastFile->isChecked() || m_watcher->m_watchListFile.isEmpty())
+  {
+    SConfig::getInstance().setWatchModel(m_watcher->saveWatchModel());
+  }
   SConfig::getInstance().setMainWindowGeometry(saveGeometry());
   SConfig::getInstance().setMainWindowState(saveState());
 
-  if (!SConfig::getInstance().ownsSettingsFile())
+  // Give the user a chance to save the ephemeral watch model or if they are using auto-loading of a
+  // file rather than the stored watchmodel
+  if (!SConfig::getInstance().ownsSettingsFile() || m_actAutoloadLastFile->isChecked())
   {
-    // Give the user a chance to save the ephemeral watch model.
     if (!m_watcher->warnIfUnsavedChanges())
     {
       event->setAccepted(false);
--- a/Source/GUI/MemWatcher/MemWatchWidget.cpp
+++ b/Source/GUI/MemWatcher/MemWatchWidget.cpp
@@ -739,11 +739,19 @@ QTimer* MemWatchWidget::getFreezeTimer() const
   return m_freezeTimer;
 }
 
-void MemWatchWidget::openWatchFile()
+void MemWatchWidget::openWatchFile(const QString& fileName)
 {
-  QString fileName = QFileDialog::getOpenFileName(this, "Open watch list", m_watchListFile,
-                                                  "Dolphin memory watches file (*.dmw)");
-  if (fileName != "")
+  QString srcFileName;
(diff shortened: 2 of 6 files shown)


dreamsyntax:
Regarding:
```
[ ] // TODO: Potentially restoreWatchModel first, then openWatchFile in autoload scenario. If they don't match, warn/prompt unsaved changes in model.
[ ] Do we want auto-load by default
```

My initial thought on the easiest and 'safe' approach to not disrupt user flows is to *always* load the stored watchlist from the config first, then load the last saved file. If they don't match 1:1 a warning is shown giving the user a chance to save the scratch, or alternatively load it instead of the file.

This works for single users working on a DMW list. However, once you introduce a shared DMW in a git repo with a team working on it, the flow can get annoying.

1. User A and User B pull the current DMW; Both are using auto-load mode in DME.
2. User A makes changes, and commits changes.
3. User B had been using the prior commit DMW, but did not make any changes. They close DMW, and later pull the new commit.
4. A few weeks go by, etc, User B has not opened DME, forgetting the above.
5. User B opens DMW. The autoload will occur and a mismatch of the config stored watchlist and the DMW will result in the warning/popup that the file does not match the scratch stored watchlist.

So long as the user 'knows' they had nothing, they could simplify tell the popup to discard and load from the file.
I don't consider this flow fully ideal, but I'm not sure what else can be done.

dreamsyntax:
I've mulled it over a bit.

I think for now, I want to disable this feature by default, and that way I can ignore doing the compare with the scratch/temp watchlist. As users decide which they prefer, they can safely migrate to the feature being enabled if that works best for them.

This also solves the multi user situation since the scratch would be *ignored* completely, so user B would not receive the false pop up in the above scenario.

Source/GUI/MainWindow.cpp
```diff
@@ -41,7 +41,17 @@ MainWindow::MainWindow()
   GUICommon::changeApplicationStyle(
       static_cast<GUICommon::ApplicationStyle>(SConfig::getInstance().getTheme()));
 
-  m_watcher->restoreWatchModel(SConfig::getInstance().getWatchModel());
+  m_actAutoloadLastFile->setChecked(SConfig::getInstance().getAutoloadLastFile());
+
+  if (m_actAutoloadLastFile->isChecked() && SConfig::getInstance().getLastLoadedFile() != "")
```
cristian64, line 46: Nit:

```suggestion
  if (m_actAutoloadLastFile->isChecked() && !SConfig::getInstance().getLastLoadedFile().isEmpty())
```

Source/GUI/MainWindow.cpp
```diff
@@ -334,6 +351,19 @@ void MainWindow::onUnhook()
...
+{
+  if (!checked)
+  {
+    if (!m_watcher->warnIfUnsavedChanges())
+    {
+      return;
+    }
+    m_watcher->clearWatchList();
+    m_watcher->restoreWatchModel(SConfig::getInstance().getWatchModel());
+  }
```
cristian64, line 364: Nit: For consistency, the first check could be an early out too.

Source/GUI/MainWindow.cpp
```diff
@@ -384,7 +414,7 @@ void MainWindow::onSplitterMoved(const int pos, const int index)
 void MainWindow::onOpenWatchFile()
 {
   if (m_watcher->warnIfUnsavedChanges())
-    m_watcher->openWatchFile();
+    m_watcher->openWatchFile(QString(""));
```
cristian64, line 417: Nit: The use of `""` in superfluous:
```suggestion
    m_watcher->openWatchFile(QString());
```

Source/GUI/MemWatcher/MemWatchWidget.cpp
```diff
@@ -739,10 +739,11 @@ QTimer* MemWatchWidget::getFreezeTimer() const
   return m_freezeTimer;
 }
 
-void MemWatchWidget::openWatchFile()
+void MemWatchWidget::openWatchFile(QString fileName)
 {
-  QString fileName = QFileDialog::getOpenFileName(this, "Open watch list", m_watchListFile,
-                                                  "Dolphin memory watches file (*.dmw)");
+  if (fileName == "")
```
cristian64, line 744:
```suggestion
  if (fileName.isEmpty())
```

Source/GUI/MemWatcher/MemWatchWidget.cpp
```diff
@@ -842,6 +843,7 @@ bool MemWatchWidget::saveAsWatchFile()
 
 void MemWatchWidget::clearWatchList()
 {
+  m_watchListFile = "";
```
cristian64, line 846:
```suggestion
  m_watchListFile.clear();
```

Source/GUI/MemWatcher/MemWatchWidget.h
```diff
@@ -33,7 +33,7 @@ class MemWatchWidget : public QWidget
   void onRowsInserted(const QModelIndex& parent, int first, int last);
   void onCollapsed(const QModelIndex& index);
   void onExpanded(const QModelIndex& index);
-  void openWatchFile();
+  void openWatchFile(QString fileName);
```
cristian64, line 36: For correctness, this should be a const reference, but I see why you've opted for passing it by value (the variable is reassigned in the definition).

If you want to spend any more time on this, a potential solution could be:

**Declaration**
```cpp
void openWatchFile(const QString& fileName);
```

**Definition**
```cpp
void MemWatchWidget::openWatchFile(const QString& fileName)
{
  QString srcFileName;
  if (fileName.isEmpty())
  {
    srcFileName = QFileDialog::getOpenFileName(...);
  }
  else
  {
    srcFileName = fileName;
  }

  ...
}
```

Source/GUI/MemWatcher/MemWatchWidget.h
```diff
@@ -49,6 +49,7 @@ class MemWatchWidget : public QWidget
   bool warnIfUnsavedChanges();
   void restoreWatchModel(const QString& json);
   QString saveWatchModel();
+  QString m_watchListFile = "";
```
cristian64, line 52:
```suggestion
  QString m_watchListFile;
```

Source/GUI/Settings/SConfig.cpp
```diff
@@ -203,6 +203,26 @@ void SConfig::setMEM2Size(const u32 mem2SizeReal)
...
+}
+
+void SConfig::setAutoloadLastFile(const bool enabled)
+{
+  setValue("autoloadLastFile", enabled);
+}
+
+QString SConfig::getLastLoadedFile() const
+{
+  return value("lastLoadedFile", "").toString();
```
cristian64, line 218:
```suggestion
  return value("lastLoadedFile", QString{}).toString();
```

Source/GUI/Settings/SConfig.h
```diff
@@ -64,6 +64,11 @@ class SConfig
 
   bool ownsSettingsFile() const;
 
+  bool getAutoloadLastFile() const;
+  void setAutoloadLastFile(bool enabled);
+  QString getLastLoadedFile() const;
+  void setLastLoadedFile(const QString fileName);
```
cristian64, line 70: Another case where a const reference should be used:

```suggestion
  void setLastLoadedFile(const QString& fileName);
```

Source/GUI/MemWatcher/MemWatchWidget.h
dreamsyntax: Is this safe to do? Is an uninitialized QString implicitly ""?

cristian64: As per [`QString::QString()`](https://doc.qt.io/qt-6/qstring.html#QString), it is safe.

Source/GUI/MainWindow.cpp
dreamsyntax: Done

Source/GUI/MainWindow.cpp
dreamsyntax: I don't understand what you mean by this

Source/GUI/MainWindow.cpp
dreamsyntax: Done

Source/GUI/MemWatcher/MemWatchWidget.cpp
dreamsyntax: Done

Source/GUI/MemWatcher/MemWatchWidget.cpp
dreamsyntax: Done

Source/GUI/MemWatcher/MemWatchWidget.h
dreamsyntax: Done

Source/GUI/Settings/SConfig.cpp
dreamsyntax: Done

Source/GUI/Settings/SConfig.h
dreamsyntax: Done

Source/GUI/MemWatcher/MemWatchWidget.h
dreamsyntax: Is this something that optimizes the code when compiles? Just for my understanding.
Made this change.

Source/GUI/MainWindow.cpp
dreamsyntax: Oh... Do you mean flipping this, so if checked == true, return? That way the inner code is not nested?

dreamsyntax: I went with this, pretty sure this is what you meant.
```c++
{
  if (checked)
    return;
  if (!m_watcher->warnIfUnsavedChanges())
    return;
  m_watcher->clearWatchList();
  m_watcher->restoreWatchModel(SConfig::getInstance().getWatchModel());
}
````

cristian64: Yes, exactly. Quite a nitpick, though.

Source/GUI/MemWatcher/MemWatchWidget.h
cristian64: I doubt there is any relevant optimization here. However, _in principle_, one would try to avoid passing objects by value, as that means a copy of the object is made. I think passing by value is only preferable if:
- The input object is cheat to copy (for example, an integer or a float, or a small structure).
- The input object will always be moved to another variable (for example, imagine a large vector that is stored in a member variable).

In this case, `QString` almost fits in the first category, as it uses implicit sharing and making copies of a `QString` is almost as cheap as copying a pointer. Although, because it has to deal with atomics to guarantee thread-safety, one should still avoid passing the object by value if possible.

...but really we don't really care about the marginal performance gain here; it was mostly about correctness.

cristian64 (approved):

PR #187, opened 2025-02-23
Lint: Use Clang-Format 19

dreamsyntax:
I'm unsure if I also need to manually update:
```
Source/.clang-format
Source/.clang-tidy
```


cristian64:
> I'm unsure if I also need to manually update:
> 
> ```
> Source/.clang-format
> Source/.clang-tidy
> ```

I'm not sure if there is any new knob in Clang-Format 19 that wasn't available in older versions and that we want to customized now, so probably the answer to your question is "no need to update the files".

cristian64 (approved):

PR #188, opened 2025-02-23
Externals: Qt6.8.2 for Windows

dreamsyntax:
* Upgrades our Qt for Windows external from 6.5.3 to 6.8.2
* This is mainly done since we may do the same for Dolphin, and I don't see any issues in my brief testing.

The only notable changes I've seen:
* Depending on your color scheme fusion dark has some tweaks.
* Window sizes are slightly different. About is larger, and Memory Viewer is smaller

The enable/disable look, drag and drop, and general usage seems just like 6.5.3, unlike the disaster the 6.7.0 experiment was (broken enable/disable colors and more).

Left is Qt6.5.3, Right is Qt6.8.2
image

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,7 +39,7 @@ jobs:
 
       - name: Build
         run: |
-          cmake Source -B .\Source\bin -A x64 "-DCMAKE_PREFIX_PATH=..\Externals\Qt\Qt6.5.3\x64"
+          cmake Source -B .\Source\bin -A x64 "-DCMAKE_PREFIX_PATH=..\Externals\Qt\Qt6.8.2\x64"
           cmake --build .\Source\bin --config ${{ matrix.configuration }} --parallel
         shell: powershell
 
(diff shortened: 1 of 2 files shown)


cristian64 (approved):
I guess Qt 6.2.4 is still used on Linux. No real hurry to upgrade, though. I'd rather the tool remains compatible with older systems than to upgrade for no real benefit on Linux.

dreamsyntax:
> I guess Qt 6.2.4 is still used on Linux. No real hurry to upgrade, though. I'd rather the tool remains compatible with older systems than to upgrade for no real benefit on Linux.

Does it not just use the System Qt version? If you are referring to the AppImage, I think we could safely target 6.4.2 at least since that's what Debian Stable ships.

cristian64:
> > I guess Qt 6.2.4 is still used on Linux. No real hurry to upgrade, though. I'd rather the tool remains compatible with older systems than to upgrade for no real benefit on Linux.
> 
> Does it not just use the System Qt version? [...]

The major/minor version will be fixed at build time. On Ubuntu 22.04, that'd be Qt 6.2. The AppImage is still bound to that limitation too (it will bundle Qt 6.2 and nothing newer).

dreamsyntax:
I just went and verified myself as well that both builds (appimage and ubuntu zip) are 6.2.4. I'll open an issue for the future.

Issue #191, opened 2025-02-26
Scanning Fix/QoL: Add Decimal Toggle When Scanning for Floats


KingJigglypuff:
I recently switched to the latest Dolphin Memory Engine after using an older version for an unspecified period of time, and one thing I really don't like is how the scanner only uses hex when scanning for floats. For instance, if I wanted to search for 2.5, I'd have to convert it into hex (2.5 in hex is 40200000) and search for that. Because if you use a decimal value when scanning, it'll think you're using hexadecimal integers (using 2.5 as an example again, putting in 2.5 will make it think you're looking for 2). This forces you to constantly convert your floats into hex whenever you want to search for a float, which I find rather unintuitive.

I don't know whether or not it's a regression, or if it's intentional. Though regardless, I'd like to suggest/request the same toggle that's used when searching for bytes also be integrated when scanning for floats (the one that lets you choose between hexadecimal, decimal, octal, or binary). Because I'm sure there's some use cases for searching floats with hexadecimal values.

dreamsyntax:
>  using 2.5 as an example again, putting in 2.5 will make it think you're looking for 2
Can you provide your configuration file?
This is not the behavior I see. Float searching works just fine:

![Image](https://github.com/user-attachments/assets/d36f528a-1bd7-44a8-b297-e005063769a9)

If you meant the watchlist, you can view it in any format you desire:
![Image](https://github.com/user-attachments/assets/7c0926d3-e5b3-4f04-a49f-91fe6d041f0e)

dreamsyntax:
Update, no need to provide a config.
There is a bug somewhere.

If you search for a byte with the radio button set to Hexadecimal and then switch back to float, it is affecting the search feature.

dreamsyntax:
Opting to just enable these views for float and double. I see no reason to restrict them.
Thanks for reporting this.

dreamsyntax:
This is now released in [v1.2.6](https://github.com/aldelaro5/dolphin-memory-engine/releases/tag/1.2.6)

PR #192, opened 2025-02-26
Bugfix: Allow selecting scanning options for float and double

dreamsyntax:
Resolves https://github.com/aldelaro5/dolphin-memory-engine/issues/191

Previously set value while viewing binary/word etc would influence scan.
I verified unaligned option also works, though doing so makes scanning take longer.

I don't really see why you'd want to use the other types, but I suppose it doesn't hurt to expose it either rather than forcing decimal.


dreamsyntax:
@KingJigglypuff try using this build.

cristian64 (approved):

Issue #193, opened 2025-02-26
Feature: Setting to collapse all Expanded Groups on save


dreamsyntax:
While the feature to save the expand/collapsed state is very useful when working in the temporary/scratch space or working solo, it can be irritating to have to manually close all Groups after a long session if committing the changes to a shared group repository.

* Add a persisting option to enable/disable removing all Group `expanded` state on save.

```
"expanded": true,
```
-> removed from .dmw when enabled

cristian64:
How about an option to expand or collapse all _recursively_?

dreamsyntax:
> How about an option to expand or collapse all _recursively_?

I guess that could be a nice feature but I personally wouldn't use it.
My change only would be for collapsing when writing the file, not affecting the actual view of DME until next open.

PR #194, opened 2025-02-27
Add Setting to collapse all Expanded Groups on save

dreamsyntax:
Implements https://github.com/aldelaro5/dolphin-memory-engine/issues/193

* Collapses expanded groups on save (to the file) - not on the immediate view.
* Only reopening the file (or closing DME and reopening) would show the changes.

Mainly intended for those storing dmws in a git repo with lots of groups to avoid unnecessarily large diffs.

image

At some point I will probably add a new Menu bar sub section, or move these to Settings. But for now since they are File operation related, I think its fine.

### DIFF
--- a/Source/GUI/MemWatcher/MemWatchModel.cpp
+++ b/Source/GUI/MemWatcher/MemWatchModel.cpp
@@ -14,6 +14,7 @@
 #include "../../Common/CommonUtils.h"
 #include "../../DolphinProcess/DolphinAccessor.h"
 #include "../GUICommon.h"
+#include "../Settings/SConfig.h"
 
 namespace
 {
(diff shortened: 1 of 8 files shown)


Source/GUI/MemWatcher/MemWatchModel.cpp
```diff
@@ -666,7 +667,7 @@ MemWatchModel::CTParsingErrors MemWatchModel::importRootFromCTFile(QFile* const
 
 void MemWatchModel::writeRootToJsonRecursive(QJsonObject& json) const
 {
-  m_rootNode->writeToJson(json);
+  m_rootNode->writeToJson(json, !SConfig::getInstance().getCollapseGroupsOnSave());
```
dreamsyntax, line 670: I don't like that I had to use SConfig here, but I couldn't think of a cleaner solution

Source/GUI/MemWatcher/MemWatchWidget.cpp
```diff
@@ -404,7 +404,7 @@ void MemWatchWidget::copySelectedWatchesToClipBoard()
 
       rootNodeCopy.appendChild(childNode);  // Borrow node temporarily.
     }
-    rootNodeCopy.writeToJson(jsonNode);
+    rootNodeCopy.writeToJson(jsonNode, true);
```
dreamsyntax, line 407: For Copy/Paste we always want to include the full json

Issue #214, opened 2025-07-24
[1.3] Unhelpful information in Value column for structs


malleoz:
This is perhaps just a matter of personal preference, but I don't gain anything from knowing that `EGG::Vector3f definition loaded`. This makes it a bit harder for me to focus on actual values. I think it's maybe helpful to be informed of the struct's name, but I feel like I'd rather have that in the "Type" column rather than "Struct".

<img width="716" height="832" alt="Image" src="https://github.com/user-attachments/assets/ce345db0-ab42-43ea-b359-3b4508c61a8b" />

dreamsyntax:
I agree it should be something else, but what should it be? Since the struct itself wont be defined

dreamsyntax:
Considering 'struct' represents a group more or less, I don't really see a good alternative without allowing user defined ToString for the struct.

For now will be dropping " definition loaded" as it just takes up room.
It would be neat to have it in the Type column, but I think the effort may make it not worth it.

dreamsyntax:
Doing more thinking about it, I think having no value (unless there is an issue) and somehow overwriting the Type as you suggested would be nicer to 'see' the real values.

PR #219, opened 2025-07-28
Show Struct name in `Type` field and show no value for Struct itself if valid

dreamsyntax:
Resolves https://github.com/aldelaro5/dolphin-memory-engine/issues/214

- 'Value' field will be blank if there is no issues resolving the struct
- 'Type' field will be the name of the struct, or just the 'Struct' string from getStringFromType if there are issues retrieving it.


dreamsyntax:
Ended up thinking about it more, instead doing what was suggested directly as it wasn't too bad of a change.
image

dreamsyntax:
@malleoz Can you give this a try and let me know if it is as you expect?

Issue #224, opened 2025-07-28
[1.3] Crash when deleting a struct


malleoz:
I'm not quite able to narrow this down to a minimal reproducible, so apologies for the large watch file.

Steps to reproduce:
- Open attached watch file
- Go to View -> Struct Editor
- Expand Kart
- Delete KartMove
- DME crashes

Callstack:
```
QArrayDataPointer<FieldDef *>::QArrayDataPointer<FieldDef *>(const QArrayDataPointer<FieldDef *> & other) Line 33
StructDef::getFields() Line 55
MemWatchModel::updateStructNode(MemWatchTreeNode * node) Line 908
MemWatchModel::updateStructEntries(QString structName) Line 902
MemWatchModel::onStructDefAddRemove(QString structName, StructDef * structDef) Line 891
MemWatchWidget::onStructDefAddRemove(QString structName, StructDef * structDef) Line 1079
QtPrivate::FunctorCall<QtPrivate::IndexesList<0,1>,QtPrivate::List<QString,StructDef *>,void,void (__cdecl MemWatchWidget::*)(QString,StructDef *)>::call(void(MemWatchWidget::*)(QString, StructDef *) f, MemWatchWidget * o, void * * arg) Line 137
QtPrivate::FunctionPointer<void (__cdecl MemWatchWidget::*)(QString,StructDef *)>::call<QtPrivate::List<QString,StructDef *>,void>(void(MemWatchWidget::*)(QString, StructDef *) f, MemWatchWidget * o, void * * arg) Line 174
QtPrivate::QSlotObject<void (__cdecl MemWatchWidget::*)(QString,StructDef *),QtPrivate::List<QString,StructDef *>,void>::impl(int which, QtPrivate::QSlotObjectBase * this_, QObject * r, void * * a, bool * ret) Line 432
StructEditorWidget::structAddedRemoved(QString _t1, StructDef * _t2) Line 224
StructEditorWidget::onDeleteNodes() Line 992
QtPrivate::FunctorCall<QtPrivate::IndexesList<>,QtPrivate::List<>,void,void (__cdecl StructEditorWidget::*)(void)>::call(void(StructEditorWidget::*)() f, StructEditorWidget * o, void * * arg) Line 137
QtPrivate::FunctionPointer<void (__cdecl StructEditorWidget::*)(void)>::call<QtPrivate::List<>,void>(void(StructEditorWidget::*)() f, StructEditorWidget * o, void * * arg) Line 174
[...]

PR #230, opened 2025-07-29
Fix MemWatchModel bug when updating structs

### LINKED ISSUE #227: [1.3] Modifying a struct causes it to no longer be expandable in watch list
Steps to reproduce:
- Open attached watch file
- Notice that `KartDynamics` can be expanded to view all members in watchlist
- Go to View > Struct Editor
- Edit `KartDynamics` by renaming a member. For example, rename `pos` to `test123`
- Press Save button
- Notice `KartDynamics` is no longer expandable in watch list. Saving and opening the watch file again allows you to expand `KartDynamics`.

[MinimalReproducible.zip](https://github.com/user-attachments/files/21476102/MinimalReproducible.zip)

malleoz:
This fixes #227. We were accidentally deleting child nodes due to a regression introduced when fixing issue #224. This is the correct expression this time. Sorry!

### DIFF
--- a/Source/GUI/MemWatcher/MemWatchModel.cpp
+++ b/Source/GUI/MemWatcher/MemWatchModel.cpp
@@ -910,8 +910,9 @@ void MemWatchModel::updateStructEntries(const QString structName)
 
 void MemWatchModel::updateStructNode(MemWatchTreeNode* node)
 {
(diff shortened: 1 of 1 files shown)


dreamsyntax (approved):
Thanks. You probably have the most usage out of this feature, so more or less trusting your findings.

Code looks fine, untested.

PR #239, opened 2025-10-15
Linux: add .dolphin-emu-wr detection, update README Troubleshooting

dreamsyntax:
Co-author: @Sewer56 
* Add .dolphin-emu-wr as another default name (nixOS)
* Add extra troubleshooting for Linux

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,8 @@ Options:
                                                      used (e.g. "Dolphin.exe" on
                                                      Windows, or "dolphin-emu"
                                                      on Linux or macOS).
+                                                     Check Task Manager
+                                                     or btop if in doubt.
 ```
 
 ## Troubleshooting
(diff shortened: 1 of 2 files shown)


dreamsyntax:
> It's the kernel 'command' name we query, not the actual program/process name 

Do you expect something like: in the help output? I think it's kind of confusing personally.
```diff
  -d, --dolphin-process-name <dolphin_process_name>  Specify custom name for
                                                     the Dolphin Emulator
                                                     process. By default,
-                                                     -platform-specific names are
+                                                    platform-specific commands are
                                                     used (e.g. "Dolphin.exe" on
                                                     Windows, or "dolphin-emu"
                                                     on Linux or macOS).
```

Sewer56:
Nah, just say `check task manager` or `btop`, or something of the like if really needs be.

PR #250, opened 2026-01-09
Don't update collapsed/invisible entries

malleoz:
If a watch group is collapsed, we shouldn't waste CPU cycles updating these entries' values, since they cannot be seen by the user. For my system, when using a watch file containing >12,000 watches (Zephiles ported an existing Paper Mario: TTYD Cheat Engine watch file to DME with a lot of data points), this drops my peak CPU usage for DME from 4.1% to 2.7%.

`freezeNodeValueRecursive` is the next hotpath in terms of CPU usage, but obviously we want freezes to persist even if you collapse groups. I'd like to explore caching locked entries rather than performing entire tree traversals on every timer expiring, but that will be a bit more complex so I want to tackle that in a separate PR.


dreamsyntax (approved):
tested on linux

Issue #241, opened 2025-10-25
Feature: Change Display of Value Type in Memory Viewer


BlakeStinson:
It would be nice if, akin to Cheat Engine, you could view values in the memory viewer as a certain data type rather than just as hexadecimal values. For instance: viewing every 4 bytes as individual floating point values in the memory viewer.

This tends to seed up finding floating point values of interest in certain structs for me when using Cheat Engine, and I think Dolphin Memory Engine would also benefit greatly from this feature.

ModSault:
I should be able to complete this in the coming week. So far I only have displaying it working, nothing with setting its value yet.

Edit: Removed image showing example of assembly code view

PR #253, opened 2026-01-13
Memory Viewer - Adjusting Display

ModSault:
Solves #241 and does what you would expect from the title

Also for context I only started committing once I was done. So if it looks like I worked in a weird order that would explain it.

# Memory Viewer - Adjusting Display

image

Right click and a new menu will appear allowing you to change the display in a similar to the memory watch. Click whatever you want the display to be. The separator will automatically increase in size if needed.

## What to test

- Highlight color. Ensure its blue if any relevant byte is selected and red if any relevant byte is modified
- The carrot position (the line showing where you'll type next). I had to rewrite it to work for larger than 2 character hex boxes. It should work (read bugs for a problem) and let you type in anywhere on the box
- Ensure the "error" sound plays when typing invalid characters in the hex area. So no 'F' key while in binary mode and such
- "Add a watch to this address", make sure it grabs the correct address. Logic for hex area and ascii area are different so test both
- "Add selection as array of bytes", ensure it grabs all 16 bytes if selecting from hex area. Ensure it doesn't do that in ascii area
-  "Edit all selection", ensure the text describes the displayed data type correctly and make sure it updates all highlighted areas. Ensure it also spits out errors correctly. Logic is the same in hex area and ascii area
- General checks. Like making sure all 16 addresses in one row are displayed correctly when changing data types. Ensure the floating point values shown are correct, ensure the assembly matches the dolphin debugger etc...

## Bugs

- Adjusting the zoom can cause characters and the carrot to appear in the wrong place. Since they are integer based it would be a long fix. If you encounter it just zoom in or out until its no longer a problem
- Typing in "522" into a one byte integer causes it to overflow. Not a big deal but be aware of that

## Other

Lemme know if you want more clarifications or screenshots somewhere

### DIFF
--- a/Source/GUI/MemViewer/MemViewer.cpp
+++ b/Source/GUI/MemViewer/MemViewer.cpp
@@ -505,9 +505,8 @@ void MemViewer::updateFontSize()
 
   m_charWidthEm = fontMetrics().horizontalAdvance(QLatin1Char('M'));
   m_charHeight = fontMetrics().height();
-  m_hexAreaWidth = (m_numColumns / m_sizeOfType) *
-                   (m_charWidthEm * m_digitsPerBox +
-                    m_charWidthEm / 2);
+  m_hexAreaWidth =
+      (m_numColumns / m_sizeOfType) * (m_charWidthEm * m_digitsPerBox + m_charWidthEm / 2);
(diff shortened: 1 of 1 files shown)


dreamsyntax (approved):
Creating a preview release for this

Issue #222, opened 2025-07-28
Feature: Container to dereference N elements of a pointer (treating as an array)


malleoz:
This is really niche use-case, but it's one I run into very often!

In Mario Kart Wii, each course has a certain number of "objects" and an array of those objects. Thus, this array varies in length on a course-by-course basis. One of my frequent use-cases nowadays is to try to capture information from the base class that all of these objects share. Since this array is variable length, I usually have to add a new watch for every object that I want to monitor. It would be a quality-of-life improvement if it were possible to do the following:
- New Container type called an Array
- When adding a watch, you can specify if it's a pointer. Add support to also specify if we are pointing to an array
- Specify the number of elements in the array
- The watchlist could then show this container, allow the user to expand, and we can see each element in the array

As an example, let's say I have the following struct:
```
struct Vector3f {
    float x;
    float y;
    float z;
};
```

Let's say there is an array `arr` of length `n` containing `Vector3f` structs at the pointer chain `0x809C18F8 -> 0x20 -> 0x0` to represent the position of `n` players. Right now, if I add this as a watch, then it will only access `arr[0]`, but I want to show `arr[0] ... arr[n]`. I could add a watch whose type is `Struct -> Vector3f`, check the `This is a pointer` box, maybe check another box that says `This is an array`, and specify a length of `n`. The number of players in a race can vary (and in my original use-case the number of objects can vary), so the ability to specify this `n` by reading from another address would be really beneficial.

This idea isn't fully thought out, so please feel free to share any thoughts! Even ignoring the dynamic `n` reading at the end, I think array pointer support could be really helpful.

PR #254, opened 2026-01-15
Adding a generic array type

jahorta:
This pull request is to add a generic array type (might be a start for #222 ). It is a container type that stores a base MemWatchEntry and uses it to determine the size of each array entry, then sets the appropriate base address for each entry in the array using that size.

image

A button has been added to the edit watch dialog to setup contents.

image

This opens a new dialog to edit the array content MemWatchEntry.

image

I've been using it for a about a month while fixing some bugs, so I believe it is relatively stable.

### DIFF
--- a/Source/Common/MemoryCommon.h
+++ b/Source/Common/MemoryCommon.h
@@ -37,6 +37,7 @@ enum class MemType
   type_struct,
   type_ppc,
   type_doubleword,
+  type_array,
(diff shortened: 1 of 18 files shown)


dreamsyntax (approved):
You can crash if you change an existing watchlist item to Array while unhooked, then expanding it - but its such a niche scenario (that shouldn't be done anyway) so approved.

Issue #249, opened 2026-01-08
Feature: Memory Viewer larger view


Sylk:
Memory viewer is currently very small and having to scroll through every line is a real PITA. If I could have it be longer and wider and have support for multi-column this would be incredible. 

The main request though is that it's able to display more lines while on screen.

<img width="776" height="478" alt="Image" src="https://github.com/user-attachments/assets/b19305af-d7dc-4ae2-828e-d3ea440d2369" />

dreamsyntax:
Very much agree, when I used this a lot I felt the same. Never got around to it, and I personally will not be.


+1 for anyone to give this PR a go

malleoz:
> // Should be a multiple of 16, or the header doesn't make much sense

For general hex-editing I use HxD, which allows you to set any number of columns. It just has to update the row labels to increment based off the number of rows. For example (`m_numColumns = 17` in this scenario):

<img width="498" height="269" alt="Image" src="https://github.com/user-attachments/assets/c94afd74-b5c0-4067-a0c9-ef3c987268af" />

Why would someone want such a strange number of columns? I have no idea. But I just figured I would mention this existing functionality in case anyone feels that we could match this behavior for the DME memory viewer.

dreamsyntax:
Thanks for implementing this @ModSault 
Released in 2026.01.24

PR #258, opened 2026-01-24
Memory Viewer; Adjust Size by Dragging Window Borders

ModSault:
# Memory Viewer; Adjust Size by Dragging Window Borders

Does as the title suggests. When enough space is present a new column/row will appear. Nothing should be hidden by the window being too small. All sizes work.

Solves #249

## What to test and look out for

- Making sure values shown are accurate (I made no adjustments to this, just changing the number of rows and columns made everything work just fine from my testing)
- Changing data types has no problem on the display
- Ensure nothing is hidden by the window. This usually means the last ASCII character is always visible. Logic is different for Double and Doubleword because they are 8 bytes in length and the "Text (ASCII)" part of the memory viewer requires a minimum of 12 characters to fit the "Text (ASCII)" text
- Adjusting zoom has it resize automatically
- Vertical Separators work properly
- Window size can't be smaller than properly showing 1 row and column no matter what

## Other

It's interesting to finally see an entire player struct at once. Makes it seem much smaller than what it was before with all the scrolling.


dreamsyntax (approved):
AppImage tested. Great job works great.

Sylk:
Thank you so much!

PR #264, opened 2026-03-28
Automatic detection of the memory size.

cristian64:
Previously, settings were available that would allow the user to specify the same simulated memory size that had been configured in Dolphin. Failing to do this precisely would lead to Dolphin Memory Engine not being able to hook into Dolphin.

The settings have been removed from the GUI now and, instead, there is an attempt to parse the simulated memory size from the data structures present in the Dolphin OS headers.

Reference for the data structures:

- https://hitmen.c02.at/files/yagcd/yagcd/chap4.html
- https://wiibrew.org/wiki/Memory_map
- https://wiibrew.org/wiki/Wii_disc

To locate ARAM and MEM2, an unideal method that consists of some hardcoded offsets (which are hardcoded in Dolphin side too) is still in use. Ideally, a more reliable mechanism consisting of memory inspection would be used; this will implemented at a later time, once the current changes have been tested sufficiently.

### DIFF
--- a/Source/Common/CommonUtils.h
+++ b/Source/Common/CommonUtils.h
@@ -1,5 +1,10 @@
 #pragma once
 
+#include <bit>
+#include <sstream>
+#include <utility>
+#include <vector>
+
 #ifdef __linux__
 #include <byteswap.h>
 #elif _WIN32
@@ -63,6 +68,21 @@ inline u64 bSwap64(u64 data)
 }
 #endif
 
+inline u16 fromBigEndianToSystemEndian16(const u16 data)
+{
+  return std::endian::native == std::endian::big ? data : bSwap16(data);
+}
+
+inline u32 fromBigEndianToSystemEndian32(const u32 data)
+{
+  return std::endian::native == std::endian::big ? data : bSwap32(data);
+}
+
+inline u64 fromBigEndianToSystemEndian64(const u64 data)
+{
+  return std::endian::native == std::endian::big ? data : bSwap64(data);
+}
+
 constexpr u32 NextPowerOf2(u32 value)
 {
   --value;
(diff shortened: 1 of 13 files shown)


dreamsyntax:
Tested on Windows - GC games only. Tried various MEM1 settings, including when MEM1 and MEM2 were both 64M.

Mixed on removing the sliders tbh, but its been consistent so I suppose its fine.
I could see there being a checkbox in the settings "Manually specify MEM1/MEM2 size" that enables the sliders as a fallback if the auto detection is failing for some reason.

cristian64:
I guess my preference is to do more testing before merging, perhaps even putting it out as a pre-release, but remove the sliders entirely _now_.

The reason of this work is that I also have some changes for Dolphin to automatically detect extended simulated memory in games that define it in the `bi2.bin` file, and it will be important that DME can then detect it automatically too.

cristian64:
@campital Can I ask if you'd be willing to do some testing on macOS?

dreamsyntax:
I don't think you'll get a reply, but I'm fine releasing it as a pre-release as is

dreamsyntax:
@cristian64 Been using this branch and its been fine.
I see there are still TODOs in the code, do you still want to merge anyway + follow up a PR later if its necessary?

cristian64:
> @cristian64 Been using this branch and its been fine. I see there are still TODOs in the code, do you still want to merge anyway + follow up a PR later if its necessary?

The TODOs can be a follow-up. The idea would be to check the memory to find MEM2 or ARAM accurately and unequivocally.

I think having a pre-release build should allow for more testing by other people, specially on macOS.

Tomorrow I'll tidy up the commit message and open it for review officially.

dreamsyntax (approved):

PR #269, opened 2026-05-04
Update build script to support macOS Universal Binary (Apple Silicon & x86_64)

RusselProuse34:
Updated the build script to support a macOS Universal Binary. All macOS builds should now run on Apple Silicon and Intel based Macs. Changes made to build.yml. Changes may need to be made to tag-build.yml but I did not feel comfortable making those changes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - master
+  workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


dreamsyntax (approved):
Verified artifact with `https://doesitarm.com/apple-silicon-app-test/`

Thanks :+1: 

I will update the tag job as well.

PR #271, opened 2026-05-06
[BLOCKED - QT Upgrade Needed] CI: MacOS 26 runner

dreamsyntax:
* Update the macOS runner 15 -> 26 (macOS 26)

AGL framework was removed in macOS 26.
The Qt version submodule expects it. We would need to upgrade the Qt module to resolve this.

> You need Qt 6.8.4 or 6.9.2 to compile on Tahoe.

Per:
https://forum.qt.io/topic/163262/ld-framework-agl-not-found-on-macos-tahoe/6

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v6
       - uses: jidicula/clang-format-action@v4.18.0
         with:
-          clang-format-version: '19'
+          clang-format-version: '22'
           check-path: Source
 
   build-windows:
@@ -124,7 +124,7 @@ jobs:
           path: dolphin-memory-engine.AppImage
 
   build-macos:
-    runs-on: macos-15
+    runs-on: macos-26
     name: 🍎 macOS
 
     steps:
--- a/.github/workflows/tag-build.yml
+++ b/.github/workflows/tag-build.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v6
       - uses: jidicula/clang-format-action@v4.18.0
         with:
-          clang-format-version: '19'
+          clang-format-version: '22'
           check-path: Source
 
   build-windows:
@@ -143,7 +143,7 @@ jobs:
             dolphin-memory-engine-${{ github.ref_name }}-linux-x86_64-binary.tar.gz
 
   build-macos:
-    runs-on: macos-15
+    runs-on: macos-26
     name: 🍎 macOS (TAG)
 
     steps:
--- a/Source/GUI/MemViewer/MemViewer.cpp
+++ b/Source/GUI/MemViewer/MemViewer.cpp
@@ -1285,7 +1285,7 @@ void MemViewer::renderSeparatorLines(QPainter& painter) const
   painter.drawLine(m_rowHeaderWidth - m_charWidthEm / 2, 0, m_rowHeaderWidth - m_charWidthEm / 2,
                    m_columnHeaderHeight + m_hexAreaHeight);
   painter.drawLine(0, m_columnHeaderHeight,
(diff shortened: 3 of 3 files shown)


dreamsyntax:
@campital would you be willing to update your Qt submodule to 6.8.4 or newer? Latest is fine as well.
If you don't want to maintain that submodule anymore I can migrate.

dreamsyntax:
Note: moved the clang change to its own PR since its not related to the blocker

@campital another ping to see if I should look into migrating to a new external

campital:
@dreamsyntax just happened to randomly visit this repo and open this PR, what are the odds! I don't have notifications on (will fix).

I can get to this tomorrow.

dreamsyntax:
> @dreamsyntax just happened to randomly visit this repo and open this PR, what are the odds! I don't have notifications on (will fix).
> 
> I can get to this tomorrow.

No rush, poking if you still intend to do this

campital:
@dreamsyntax Yep, will submit a PR in a bit.

It seems Homebrew recently fixed the `@rpath` problems that prevented `macdeployqt` from running properly, so we no longer need to provide a standalone version as a submodule. I also switched to `qt_generate_deploy_app_script` in CMake, which generates deployment scripts on all supported platforms (should be able to replace Linux `appimage.sh` with a bit more work).

On Windows, we can use this deployment script instead of the manual DLL copying. Also, the binaries provided by Qt at https://www.qt.io/development/download-qt-installer-oss should work, so I removed the Windows Qt submodule as well (I feel like hosting binaries on GitHub is probably not best practice unless required). I'll need to find a clean way to make this work with the GitHub actions runner.

Let me know if this sounds ok - I wasn't sure if there was some other reason we maintained the submodules.

dreamsyntax:
> It seems Homebrew recently fixed the `@rpath` problems that prevented `macdeployqt` from running properly, so we no longer need to provide a standalone version as a submodule. I also switched to `qt_generate_deploy_app_script` in CMake, which generates deployment scripts on all supported platforms (should be able to replace Linux `appimage.sh` with a bit more work).

I'm fine with that for macOS, provided it doesn't bloat the binary or anything weird.
Lets keep the changes macOS only for now.

As for Linux, I'm not sure what appimage.sh has to do with this.


> On Windows [...] I wasn't sure if there was some other reason we maintained the submodules.

It's to match how dolphin-emu does it (and to strip unneeded components) and to pin to specific Qt versions. Since this is working fine and there's no immediate reason to change it I'd prefer to leave that alone. If you do pursue this for Windows, I would say do it on a separate PR from the macOS fix/update.

campital:
Unfortunately it seems `macdeployqt` is still slightly broken in the Homebrew package and creates a more bloated package as a result. I found a few other problems with `windeployqt` that don't seem trivially avoidable (copies libs into the wrong place, copies unnecessary libs), so maybe it would be best to keep the current approach. However, we could switch to `aqtinstall` for the actions workflows, which would allow us to delete the submodules and still pin specific versions (also fixing macOS). Should I do this?

> As for Linux, I'm not sure what appimage.sh has to do with this.

My idea was to unify deployment and move everything into CMake (since the new `qt_generate_deploy_app_script` would make this more natural). Creating an AppImage would only be a few more lines in CMakeLists.

dreamsyntax:
> Unfortunately it seems `macdeployqt` [...] it would be best to keep the current approach. However, we could switch to `aqtinstall` for the actions workflows, which would allow us to delete the submodules and still pin specific versions (also fixing macOS). Should I do this?


If we can isolate it just to macOS runners for now, I'm fine with it as an experiment. I would prefer to have that around a few cycles before exploring Win/Linux

campital:
Made a PR for this: https://github.com/aldelaro5/dolphin-memory-engine/pull/282.

PR #277, opened 2026-06-15
Add CMake build dir to .gitignore, added memory size config back, fix ARAM detection for GC games on Linux

xchellx:
- Added build/ to .gitnore to exclude CMake build output (as per README.md).
- Brought back the manual memory configuration with an automatic detection setting thats checked by default (for backwards compatibility with the automatic detection changes).
- Fixed ARAM detection for GameCube games (Dolphin uses an anonymous private memory map for this unlike for Wii where it is in ExRAM which is in a shared memory map). Linux only for now. I don't have a Windows nor a Mac machine to write the necessary changes for thos platforms.

The reason for bringing manual memory configuration back is that bi2.bin can lie and the BAT registers can also lie. Custom configurations of these are edge cases automatic detection fails at. Bringing back manual override for these edge cases (as an option) is the best solution here, else it'd be a fight of who is the right source of truth for how much memory there is. The automatic detection is still on by default, but can be turned off with manual size sliders as like before.

Further more, the ARAM detection being fixed only for Linux is because I do not have the necessary resources nor know-how to write for these platforms. I am sure there can be similar approaches done on Windows and Mac but I cannot write for these.

The .gitignore change is most logical as it matches the other auxiliary files from vscode and etc. that should be ignored. README.md directly mentions this directory so its only logical to exclude it (its simply CMake build output).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ Source/.idea/*
 .DS_Store
 .vs/
 .vscode/
+build/
--- a/Source/DolphinProcess/Linux/LinuxDolphinProcess.cpp
+++ b/Source/DolphinProcess/Linux/LinuxDolphinProcess.cpp
@@ -3,6 +3,7 @@
 #include "LinuxDolphinProcess.h"
 #include "../../Common/CommonUtils.h"
 #include "../../Common/MemoryCommon.h"
+#include "../../GUI/Settings/SConfig.h"
 
 #include <array>
 #include <cctype>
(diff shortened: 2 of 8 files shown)


dreamsyntax:
Thanks! Pinging @cristian64 since you were pushing for that auto detection memory related feature in Dolphin.
I don't see any issue with re-adding the manual control here as a backup.
As for the ARAM stuff, I don't know any of my games that use it off the top of my head - but the code looks fine at a glance

cristian64:
Could I ask for this to be split into three separate PRs?

1. One for fixing ARAM detection
2. One for the `.gitconfig` changes
3. One for bringing back the the sliders for manual configuration

In particular, I'd like to discuss the need for third one, as I couldn't find a case where they made sense now. Even if `bi2.bin` can lie, Dolphin Memory Engine is following the emulated memory size that Dolphin reports. Of course, if detection fails, we should rather fix that, as per the fix you are proposing the ARAM detection. (Mostly opening the discussion here; I'm not all knowing; but I'd rather we did this only if we know of a realistic use case that would not require tampering with Dolphin until it crashes or something like that.)

xchellx:
As requested, I have split the this PR into 3 individual PRs:
1. #278
2. #279
3. #280

Regarding the manual memory configuration sliders in #279: I understand the preference for avoiding manual overrides, but I have included them as a 'fail-safe' for specific edge cases (such as when bi2.bin reports memory boundaries that conflict with BAT registers). I have detailed the technical justification for this in the PR description, but I am entirely open to dropping this feature and/or further discussion in the PR. However, if there is no way forward in that path, another path forward should be looked for, in my opinion.

PR #280, opened 2026-06-16
Fix ARAM detection for GC games on Linux

xchellx:
Fixed ARAM detection for GameCube games (Dolphin uses an anonymous private memory map for this unlike for Wii where it is in ExRAM which is in a shared memory map). Linux only for now. I don't have a Windows nor a Mac machine to write the necessary changes for thos platforms.

Dolphin gives a private anonymous memory map for ARAM for GameCube games (not shared memory unlike GameCube games in Wii mode). This fixes ARAM detection by including anonymous private memory maps and checking its size and a common signature I saw across many games' ARAM dumps. This was the best solution I found with this problem as Dolphin intends for ARAM to only be accessed by its DSPManager interface which makes things difficult for ie. DME as an external tool not part of/compiled with Dolphin.

Further more, the ARAM detection being fixed only for Linux is because I do not have the necessary resources nor know-how to write this fix for Windows and Mac. I also lack machines with these operating systems. I am sure there can be similar approaches done on Windows and Mac but I cannot write for these.

See:
1. [DSP.cpp](https://github.com/dolphin-emu/dolphin/blob/1bc93fd16d5a452bedcc5437923abd0d9fcb8c52/Source/Core/Core/HW/DSP.cpp#L138)
2. [MemoryUtil.cpp](https://github.com/dolphin-emu/dolphin/blob/1bc93fd16d5a452bedcc5437923abd0d9fcb8c52/Source/Core/Common/MemoryUtil.cpp#L128)

### DIFF
--- a/Source/DolphinProcess/Linux/LinuxDolphinProcess.cpp
+++ b/Source/DolphinProcess/Linux/LinuxDolphinProcess.cpp
@@ -63,8 +63,11 @@ bool LinuxDolphinProcess::obtainEmuRAMInformations()
     std::string line;
     while (getline(mapsFile, line))
     {
-      if (line.find("/dev/shm/dolphin-emu") == std::string::npos &&
-          line.find("/dev/shm/dolphinmem") == std::string::npos)
+      if (  // Shared mappings
+          (line.find("/dev/shm/dolphin-emu") == std::string::npos &&
+           line.find("/dev/shm/dolphinmem") == std::string::npos) &&
+          // Anonymous mappings
+          (line.find("rw-p") == std::string::npos && line.find("00:00 0") == std::string::npos))
         continue;
 
       const std::vector<std::string> lineParts{Common::splitBySpace(line)};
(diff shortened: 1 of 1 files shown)


cristian64:
It seems the branch currently includes the changes in #279. Could we leave in this branch only the changes for the ARAM so that they can be looked at in isolation?

(I'd like to focus on this change before we discuss brining the config back.)

xchellx:
Done. The commits are isolated per PR.

cristian64:
How could I verify that the change is working? Is there any particular game that is needed?

xchellx:
Metroid Prime and Metroid Prime 2: Echoes are games that I know make use of ARAM for streaming data from PAK files. This is what I primarily tested with but I also tested this among my game library such as Paper Mario TTYD. ARAM should be allocated by Dolphin regardless if the game makes actual use of it. But games that do at least initialize ARAM usually have `02 9F 00 10` at the beginning of ARAM, so that can help verify if you are looking at ARAM. Games that dont initialize ARAM will just have 0s.

As for verifying if its working, the "Go To ARAM" button should be clickable and of course youd see `02 9F 00 10` at the beginning for most games (some games may be just 0s). You can enable Debug UI for Dolphin and View -> Memory and then Export -> Dump ARAM and open `/Dump/aram.raw` (in your dolphin /User directory or ~/.local/share/dolphin-emu directory`) in a hex editor and compare its content with what DME shows for extra verification.

cristian64:
I have been able to verify that, indeed, currently ARAM shown by DME does not match the buffer exported via **Export > Dump ExRAM** in Dolphin. So, there is definitely something to address here.

(I haven't checked if this broke with my recent changes in #264, but very likely.)

> But games that do at least initialize ARAM usually have 02 9F 00 10 at the beginning of ARAM, so that can help verify if you are looking at ARAM. Games that dont initialize ARAM will just have 0s.

I remember asking about the `02 9F 00 10` numbers in https://discord.com/channels/521709831132807179/673650160978362378/1487480625832132749 (March 2026), but didn't get a response. I wanted to use those numbers to identify MEM2 with some accuracy. I didn't realize, though, that ARAM also showed the same numbers when detected correctly.

Source/DolphinProcess/Linux/LinuxDolphinProcess.cpp
```diff
@@ -140,24 +143,49 @@ bool LinuxDolphinProcess::obtainEmuRAMInformations()
   {
     const u64 size{secondAddress - firstAddress};
 
-    if (m_ARAMSize && size == Common::NextPowerOf2(m_MEM1Size) &&
-        (offset & 0x00040000) == 0x00040000)
+    if (systemInfo.isDiscMagicWordGCKnown() && systemInfo.isBootCodeKnown() &&
+        size == dolphinOSGlobals.getARAMSize())
```
cristian64, line 147: Isn't this equivalent to:

```suggestion
    if (size == m_ARAMSize)
```

`m_ARAMSize` would only be greater than zero if a GC game has been detected; no need to check the magic numbers again.

We should probably add `if (!size) continue;` to discard empty segments (although I'm not sure if they can realistically appear).

Source/DolphinProcess/Linux/LinuxDolphinProcess.cpp
```diff
@@ -140,24 +143,49 @@ bool LinuxDolphinProcess::obtainEmuRAMInformations()
...
-      m_MEM2AddressStart = firstAddress;
-      m_MEM2Present = true;
-      break;
+      // Wii uses a shared map for MEM2 and ARAM (ExRAM)
+      // TODO: Better heuristics for wii? Validating the memory for common patterns?
+      if (m_ARAMSize && size == Common::NextPowerOf2(m_MEM1Size) &&
+          (offset & 0x00040000) == 0x00040000)
+      {
+        m_emuARAMAdressStart = firstAddress;
+        m_ARAMAccessible = true;
```
cristian64, line 175: There is something around here that doesn't sound right to me. As far as I know, `m_ARAMAccessible` should never end up being `true` for Wii games. So, the inline comment above doesn't align to what I thought I knew.

For Wii games, there will be MEM1 and MEM2 (ExRAM), but no ARAM.

So, either the block is no longer needed, or the inline comment is not accurate. (I _think_ it's the former.)

Source/DolphinProcess/Linux/LinuxDolphinProcess.cpp
cristian64: On a side note, I was under the impression that the size of the memory map for the ARAM memory would not necessarily match the ARAM size that Dolphin reports (i.e. `m_ARAMSize`), but a greater buffer (the next power of 2).

dreamsyntax:
Is this abandoned?

PR #282, opened 2026-06-24
Use aqtinstall (Qt 6.11.1) for GitHub workflows on macOS

campital:
- Updated the workflow runner to macOS 26.
- Workflow now uses `aqtinstall` to install a stripped version of Qt 6.11.1 on Mac.
- Removed the `Qt-macOS` submodule.
- Now targets macOS 12.0.

In the future, we could use this approach to build on Windows and Linux as well. Held off on the other CMake changes I mentioned in https://github.com/aldelaro5/dolphin-memory-engine/pull/271#issuecomment-4777236968.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -124,28 +124,25 @@ jobs:
           path: dolphin-memory-engine.AppImage
 
   build-macos:
-    runs-on: macos-15
+    runs-on: macos-26
     name: 🍎 macOS
 
     steps:
       - name: Checkout code
         uses: actions/checkout@v6
 
-      # Current Runner has cmake we want by default; Omitting for now but restore if they stop bundling it
-      # - name: Install Dependencies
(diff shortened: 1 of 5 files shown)


dreamsyntax:
Is there a way we can check the old deployment target on the existing binary? I'm a bit ignorant on macOS related things so just curious on the previously implicit value

campital:
It seems to target the current OS version by default. I realized that after I downloaded the build artifact (macOS 26) and it wouldn't run on my system (macOS 15).

I think we should have previously been targeting macOS 12 or some other non-latest version anyways, no reason to exclude them and I don't think we need the newer OS frameworks.

dreamsyntax:
Sounds good to me, 12 it is.

dreamsyntax (approved):
LGTM. Untested (no Mac available) trusting campital

